Issue #7, opened 2017-12-21
Set up scaffolding


jellegerbrandy:
We need different directories for 

    /player
    /portal
    /assets (shared assets)

[elia does not agree with this ^^^]

We also need different `build` and `test` commands, probably along the lines of:

    yarn testPlayer   (run tests in /player)
    yarn testPortal (run tests for /portal
    yarn test (run all tests)

And, similarly:

     yarn buildPlayer
     yarn buildPortal
     yarn build

@bent0b0x , does it make sense like this?

- [ ] still need to separate build processes for embeddable player and portal

jellegerbrandy:
@eliawk , does this look reasonable to you?

eliawk:
😄 
my proposal for the scaffolding is more components oriented, something like
/components
/modules
/section 

![](http://bradfrost.com/wp-content/uploads/2013/06/atomic-design.png)

i also found this https://arc.js.org/

bent0b0x:
@jellegerbrandy what you laid out makes sense. I'll try and finish up the portal scaffold today and then once I do move it into its own directory (`/portal`). 

As for `arc.js`, I think we can incorporate most of what it includes on our own, so not sure if it gives us too much benefit at this point.

bent0b0x:
Also @eliawk I still plan on adding more structure to the `js`, including separate directories for `components` and such

jellegerbrandy:
We still need a separate builds for player and portal.

jellegerbrandy:
good points.  Here are some more observations:

- there is a difference bewteen "npm package" and "github repo".  t's ok to develop different packages in one repo (look wat web3 does: https://github.com/ethereum/web3.js/tree/1.0/packages)
- having different builds from the same code: we may also imagine that we have different player-builds (with or without certain functionality, say); that we have builds for different platforms, etc.
- (Not directly relevant to the discussion, but: I think the publsh-to-npm-and-then-import-again is just cumbersome and unnecessary, as we can also install packages directly from github.)
- **the bottom line, I think, is that** developing both player and portal in the same repo would not only be for sharing static assets, but also to (1) keep development in sync (i.e. not having to keep different repo's  in sync) and (2) share devpops infrastructure (test suite, CI, deployment procedure, etc.

At this point, my personal opinion tends to agree with @bent0b0x  and to have separate repositorie after all.

jellegerbrandy:
I'm closing this for now - we can revisit this later if necessary

Issue #8, opened 2017-12-22
Set up  basic pages for upload tool


jellegerbrandy:
there are some routes defined here with whcihw e can start: https://github.com/Paratii-Video/paratii-portal/wiki/Portal-Specs:-Upload-Tool. 

We can also set up rudimentiry pages there (while waiting for the UX design).

- [x] /uploader/upload-file
- [ ] /uploader/video/edit/{videoId}
- [ ] /uplaoder/video/status/{videoId}

jellegerbrandy:
The pages are there; they need a lot of work

Issue #6, opened 2017-12-21
Configure continuous deployment for player


jellegerbrandy:

felipegaucho:
Lmk @Leeaandrob if by any chance I'm closing this prematurely :)

PR #9, opened 2017-12-22
7 - app infrastructure

bent0b0x:
**Issue**
#7 

**TODO**
- [x] update readme
- ~~[ ] figure out approach for production deployment~~ this should not be a problem
- [x] set up actual browser interaction tests (tracking this issue in #10)

**NOTE**: when reviewing changes, you can skip past files in the `flow-typed` directory. These were pulled in by the `flow-typed` cli and help us with our static type-checking. We do not directly maintain them; no need to scrutinize them too much.

**Build System Features/Tech Stack**
- **react**
  - for UI
- **redux**
  - for application state management
  - `redux-thunk` is installed for async actions. We alternatively may want to use `redux-saga`
- **react-router** for client-side routing
- **webpack**:
  - including react/redux hot module replacement 😱 

     hmr demo
- **flow**, for static type checking (visit https://flow.org for more info)
  - this may be one of the more controversial bits of the system, as not everyone is familiar with it. However, I think it is extremely helpful in terms of codebase stability/readability. 
  - the goal should be to use `flow` extensively, but one of the great things about it is that we can turn it on on a per-file basis (by adding `/* @flow */` to the top of a file). This way we can ease into it w/o slowing down development in the near term
- **karma/jasmine/enzyme** for unit testing
- **styled-components** (visit https://www.styled-components.com/ for more info)
  - this allows us to write all of our styles in js (not to be confused w/ inline-styles, which this is not)
  - the main benefit here is that we do not need to specify class names/identifiers anymore; this is very helpful in the react/componentized world that we will be developing in
- **express** (as a simple dev server)
  - this is used in conjunction with `webpack-dev-middleware` and `webpack-hot-middleware` to enable live reload
- **eslint** for linting js
- **stylelint** for linting `styled-components` styles
- **prettier** for code formatting (I suggest installing a prettier plugin for your editor of choice, to enable auto-formatting on save. Either way, the formatting will save in a pre-commit hook but this is still nice)


.circleci/config.yml
```diff
@@ -1,30 +1,20 @@
...
-
     steps:
       - checkout
-
-      # Download and cache dependencies
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "package.json" }}
-          # fallback to using the latest cache if no exact match is found
+          - v1-dependencies-{{ checksum "yarn.lock" }}
```
bent0b0x: Since `yarn.lock` is that actual source of truth for our dependency tree, I think we should base the checksum off of this file

flow-typed/npm/babel-core_vx.x.x.js
```diff
@@ -0,0 +1,253 @@
+// flow-typed signature: 58b12a9978527dd32132602d45f30ba4
```
bent0b0x: Files in the `flow-typed` directory are type definitions for external libraries that will help us in our static type-checking

src/scripts/components/App.js
```diff
@@ -0,0 +1,51 @@
...
+  match: {
+    url: string
+  }
+};
+
+const Wrapper = styled.div`
+  text-align: center;
+`;
+
+const Header = styled.header`
```
bent0b0x: This is a good, simple example of how to use `styled-components`

src/scripts/containers/AppContainer.js
```diff
@@ -0,0 +1,9 @@
+import { connect } from "react-redux";
```
bent0b0x: This is the general structure of our connected `react/redux` containers. When we actually have data in our store and actions to dispatch, those will populate the functions `mapStateToProps` and `mapDispatchToProps`

src/scripts/createStore.js
```diff
@@ -0,0 +1,23 @@
+import { createStore, applyMiddleware, compose } from "redux";
+import thunk from "redux-thunk";
+
+import reducer from "reducers/index";
+
+const composeEnhancers = /* window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || */ compose;
```
bent0b0x: When we want to use the redux-devtools chrome extension, simply uncomment `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__`

test/index.test.js
```diff
@@ -0,0 +1,7 @@
+var Enzyme = require("enzyme");
+var Adapter = require("enzyme-adapter-react-16");
+
+Enzyme.configure({ adapter: new Adapter() });
+
+var testsContext = require.context(".", true, /.test.js$/);
+testsContext.keys().forEach(testsContext);
```
bent0b0x: This is boilerplate that allows us to bundle all of our tests using webpack, without creating a separate bundle per file.

test/test-utils/createTestRenderer.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
```
bent0b0x: This is a test util I use for testing react components, example of usage can be seen in `App.test.js`

bent0b0x:
This takes care of the basics for #6

bent0b0x:
This gets us started for #8. There is a route (`/uploader/upload-file`) that will render an `UploadFile.js` component where we can get started!

src/scripts/components/App.js
```diff
@@ -0,0 +1,51 @@
+/* @flow */
+
+import React, { Component } from "react";
+import { Route } from "react-router-dom";
+import styled from "styled-components";
+import logo from "assets/img/paratii_logo.png";
```
bent0b0x: With our `webpack` setup, images are being imported as inline data urls

jellegerbrandy:
This is great, ben, many thanks!

Issue #11, opened 2017-12-23
Do not build everything in a single bundle.js


jellegerbrandy:
At the moment, `yarn build` packages all data, including static assets like images, and source maps of code, etc, in a single `bundle.js`. 

This is not scaleable and f*ks up caching, we need an output in which we can serve assets directly (bu apache/nginx) from a single `static` folder

In other words: I liked to output of `react-scripts build` much better (cf commit d56d82f1bd66bb53ac456859446626f0fb9d73fc)

bent0b0x:
@jellegerbrandy can this be closed or are there addl. parts of the build process that you'd like to see improved?

jellegerbrandy:
Closing this for now - it is really an optimization matter, and that is something we should not bother with too much at this early stage.

PR #13, opened 2017-12-23
11 |  Bundle Improvement

bent0b0x:
**Issue**
#11 

**Work Done**
- do not add assets as inline data urls; instead, use `file-loader` to make them available in the `build/` dir
- Change sourcemap type and only generate sourcemaps in dev/test mode
- uglify code in prod build

@jellegerbrandy do you still want to generate any kind of sourcemaps in production? If so, which from [this list](https://webpack.js.org/configuration/devtool/) do you think would be best?

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "dev": "concurrently \"yarn run dev-client\" \"yarn run dev-server\"",
     "build":
       "yarn run clean && yarn run prep && NODE_ENV=production webpack --progress --colors",
-    "prep": "mkdir build/ && cp src/index.html build/",
+    "prep":
+      "mkdir build/ && cp src/index.html build/ && mkdir build/assets/ && cp -r src/assets/ build/assets/",
     "clean": "rm -rf build/",
     "lint:scripts": "eslint src test --cache",
     "lint:styles": "stylelint 'src/**/*.js'",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -7,14 +7,11 @@
   "author": "Benjamin Baum <benjaminjbaum@gmail.com>",
   "license": "MIT",
   "scripts": {
-    "dev-client":
-      "yarn run clean && yarn run prep && webpack --watch --colors --progress",
-    "dev-server":
-      "nodemon --watch src/server --watch node_modules src/server/index.js localhost:8080",
+    "dev-client": "yarn run clean && yarn run prep && webpack --watch --colors --progress",
+    "dev-server": "nodemon --watch src/server --watch node_modules src/server/index.js localhost:8080",
     "dev": "concurrently \"yarn run dev-client\" \"yarn run dev-server\"",
```
bent0b0x: ☝️ just a formatting change

package.json
```diff
@@ -24,7 +21,12 @@
...
   },
   "lint-staged": {
-    "*.js": ["prettier --write", "eslint --cache", "stylelint", "git add"]
+    "*.js": [
+      "prettier --write",
+      "eslint --cache",
+      "stylelint",
+      "git add"
+    ]
   },
```
bent0b0x: ☝️ just a formatting change

PR #19, opened 2017-12-26
12 standard linting

bent0b0x:
**Issue**
#19

**Problem**
`prettier` was overriding rules from `eslint-standard`. 

**Work Done**
Rip out `prettier` and rely on a pre-commit hook that runs `eslint --fix` to enforce formatting.


bent0b0x:
@jellegerbrandy this PR rips out `prettier` in favor of eslint standard. The major thing that we lose by removing prettier is that our codebase will no-longer enforce a code-style all the way through. `eslint-standard` gets us part of the way there but it gives more wiggle room than does `prettier` (which gives practically 0). That's my last ditch effort to save `prettier`, but feel free to merge this anyways :)

jellegerbrandy:
Ok, this is prefect. 
The only thing I did was remove the `--fix` flag from the pre-commit; it's just a personal preference to not have any side-effects when doing a commit.

PR #34, opened 2017-12-28
26 - refine test configuration

bent0b0x:
**Issue**
#26 

**Work Done**
- `jasmine => mocha/chai`
- reorganize testing scripts and test directories

@mrocha94 one question for you; on this branch, I removed the `sleep 30` from circleCI before chimp tests because I think the browser will wait to load until the webpack bundle is finished. So far I haven't seen an issue on CI. You added in the `sleep` so I just want to make sure you don't have a different reason for why it needs to be in there.


mrocha94:
There is a conflict because of 725015ef20e870c0dc7f677afeb671c2567417f0 on dev.
I think we should merge dev into this branch and PR on dev, do you agree?

bent0b0x:
@mrocha94 done!

mrocha94:
About the sleep, it's to prevent tests from failing because the server didn't start

bent0b0x:
Okay. I am going to remove the `sleep` for now because I think our tests will wait for the webpack bundle when the dev server is booting, but if it becomes a problem we will add it back in 👍

jellegerbrandy:
Seems good to me :-)

PR #35, opened 2017-12-28
33 adding parity

geckoslair:
i've added new scripts:

"dev-parity": "concurrently \"yarn parity\" \"yarn dev-client\" \"yarn dev-server\"

and

"parity": "parity --config parity-config.toml"

paraty-config.toml and chain.json are used for rightly configure the dev chain.


package.json
```diff
@@ -10,6 +10,7 @@
     "dev-client": "yarn clean && yarn prep && webpack --watch --colors --progress",
     "dev-server": "nodemon --watch src/server --watch node_modules src/server/index.js localhost:8080",
     "dev": "concurrently \"yarn dev-client\" \"yarn dev-server\"",
+    "dev-parity": "concurrently \"yarn parity\" \"yarn dev-client\" \"yarn dev-server\"",
```
bent0b0x: I would run `parity` concurrently in the `"dev"` script. We almost always will want to run parity when developing, yes?

bent0b0x:
After @jellegerbrandy 's discussion, I think the target of this PR should be changed to `dev`. Additionally, in general this looks good to me. But I don't think we should merge it until the actual debug functionality has been added to the component. If we just want to merge the `parity` stuff, then I would remove the `/debug` route from this PR

geckoslair:
rebased

package.json
geckoslair: we had a configuration like this for testrpc on meteor, sometimes it's better run it separately for better debug. Change as you prefer man! =)

bent0b0x: Nah that makes sense, 👍 !

PR #38, opened 2018-01-03
Player thus far

bent0b0x:
This is a WIP, but I don't see any harm in merging right now just so we reduce merge conflicts w/ other tasks.


webpack.config.js
```diff
@@ -90,7 +94,11 @@ var config = {
         new webpack.DefinePlugin({
           "process.env.NODE_ENV": JSON.stringify("production")
         }),
-        new webpack.optimize.UglifyJsPlugin()
+        new UglifyJsPlugin({
```
bent0b0x: This is to prevent issues with uglifying es6 code

Issue #42, opened 2018-01-06
Investigate moving back to yarn


bent0b0x:
We had to switch to `npm` via #38 , because of issues installing certain dependencies (I think it was web3?) with `yarn`. However, I was on a very old version of `yarn` that may have explained the trouble. Will investigate if we can switch back.

bent0b0x:
Someone else feel free to try, but this is not working for me, but with different issues from before.

Basically, `yarn` seems to refuse to install `tar.gz`. After `yarn` it is never present in `node_modules`. Even if you do `yarn && `yarn add tar.gz`. You end up w/ this warning:

```
warning Pattern ["tar.gz@^1.0.7"] is trying to unpack in the same destination "/Users/benjaminbaum/Library/Caches/Yarn/v1/npm-tar.gz-1.0.7-577ef2c595faaa73452ef0415fed41113212257b" as pattern ["tar.gz@^1.0.5"]. This could result in a non deterministic behavior, skipping.
```

This package is deprecated, but it should still install. I think it is related to this [yarn issue#4812](https://github.com/yarnpkg/yarn/issues/4812).

bent0b0x:
Should reinvestigate

jellegerbrandy:
I am using yarn without problems. @bent0b0x , you still have problems? The `package.json` commands use yarn.

If no-one reports problems, we should update the README and the circleci tests as well to reflect that we are using `yarn`.

Leeaandrob:
I am using yarn only to build `yarn build` on stage and production environment. @bent0b0x

bent0b0x:
FYI the question w/ `yarn` is not whether we can use it to run our build scripts, but rather whether we can use it to install our dependencies. Looking into this now!

felipegaucho:
Closing this, please undo if still relevant.

PR #45, opened 2018-01-07
42 - yarn city

bent0b0x:
Back to `yarn`!
  
#42


.flowconfig
```diff
@@ -2,6 +2,7 @@
 .*/node_modules/immutable/.*
 .*/node_modules/documentation/.*
 .*/node_modules/oboe/.*
+.*/node_modules/ipfs/test/.*
```
bent0b0x: this was causing a flow error

Issue #24, opened 2017-12-28
Migrate features/profile.test.js


jellegerbrandy:
This fiel contains the test from the meteor app in meteor player and describes the flow of wallet handling/account creation.

The work to be done is to migrate these tests one by one (or decidign to delete the test if it is not relevant anymore) and implement the corresponding functionality using `paratii-lib`.

PR #47, opened 2018-01-10
24 profile

mrocha94:
Signup:
* form

Login:
* form
* state
* cookie
* logout

Profile:
* page
* state

ref #24


src/scripts/actions/UserActions.js
```diff
@@ -0,0 +1,30 @@
...
+const loginSuccess = createAction(LOGIN_SUCCESS)
+const logoutAction = createAction(LOGOUT)
+
+const sleep = (ms) => {
+  return new Promise(resolve => setTimeout(resolve, ms))
+}
+
+export const login = (email: string, password: string) => (dispatch: Dispatch<*>) => {
+  dispatch(loginRequested())
+  sleep(200).then(() => {
```
bent0b0x: What is the use of the `sleep`? Is this not actually logging in right now and just faking it? In the future would this be replaced by a call to `paratii-lib`?

src/scripts/components/Login.js
```diff
@@ -0,0 +1,42 @@
...
+  font-size: 20px;
+`
+
+const Title = styled.header`
+  background-color: #fff;
+  height: 50px;
+  padding: 20px;
+  display: flex;
+  align-items: center;
+  color: #95989a;
```
bent0b0x: I think we should move all of our color values to a new `constants/StyleConstants.js` file, and re-create the color constants we had in sass in `paratii-player`. I'll create an issue for this

src/scripts/reducers/UserReducer.js
```diff
@@ -0,0 +1,37 @@
...
+
+const reducer = {
+  [LOGIN_REQUESTED]: (
+    state: UserRecord
+  ): UserRecord => {
+    return state
+      .set('isLoggingIn', true)
+      .set('email', null)
+      .set('name', null)
+  },
```
bent0b0x: ☝️  the logic here is fine, but it is more performant to do this in one operation:

```javascript
return state.merge({
  isLoggingIn: true,
  email: null,
  name: null
});
```

src/scripts/selectors/index.js
```diff
@@ -1,7 +1,18 @@
...
 
 import type { RootState } from 'types/ApplicationTypes'
 
+/* Videos */
 export const getVideo = (state: RootState): ?VideoRecord => state.video
+
+/* Users */
+export const getUser = (state: RootState): ?UserRecord => state.user
+export const isLoggingIn = (state: RootState): ?boolean => (state.user) ? state.user.isLoggingIn : false
+export const isLogged = (state: RootState): ?boolean => {
```
bent0b0x: Can this be renamed to `isLoggedIn`?

src/scripts/selectors/index.js
```diff
@@ -1,7 +1,18 @@
...
+  const user = state.user
+  return !!(user && user.email && !isLoggingIn(state))
```
bent0b0x: `isLogged` is a relatively simple selector, but here is the plan for any selector that is more than just reading a property off of a piece of state (and there was no way for you to know this ahead of time, so no worries):

- We will use the [reselect](https://github.com/reactjs/reselect) library to define these selectors
- This lib exports a function called `createSelector` which takes in other selectors and creates a new selector that takes the outputs of the other selectors as its inputs. And in addition, if its inputs are the same as the last time it was invoked, it will not recompute its return value and return a cached result instead.
- this takes advantage of the immutability of our store, because comparisons of whether or not the previous and current inputs are the same is just a `===` check
- also, complex selectors like this I think should go in separate modules. This one would go in `selectors/UserSelectors.js`.

So, this selector I think should look like this:

```javascript
// selectors/UserSelectors.js
import { createSelector } from 'reselect';

import { getIsLoggingIn } from 'selectors/index';

export const  getIsLoggedIn = createSelector(
  [getUser, getIsLoggingIn], (user, isLoggingIn) => user && user.get('email') && !isLoggingIn
)
```

Let me know if this makes sense!

src/scripts/selectors/index.js
```diff
@@ -1,7 +1,18 @@
...
+import UserRecord from 'records/UserRecords'
 
 import type { RootState } from 'types/ApplicationTypes'
 
+/* Videos */
 export const getVideo = (state: RootState): ?VideoRecord => state.video
+
+/* Users */
+export const getUser = (state: RootState): ?UserRecord => state.user
+export const isLoggingIn = (state: RootState): ?boolean => (state.user) ? state.user.isLoggingIn : false
```
bent0b0x: Is `?boolean` accurate? I would think this value will never be `undefined/null`. If so, you can drop the `?` (same question with selectors above and below

bent0b0x:
The two main things that I think should be addressed:

1. Avoid duplicating `styled-components`. Move shared components (e.g. `Title`) into a new directory from where any complex component can import them
1. Use `reselect` library for any selector that is more than a simple `.get()`

src/scripts/actions/UserActions.js
HugoMitsumori: It's just to fake the time response of the lib call. In the future we replace it for the call

bent0b0x: Got it

PR #53, opened 2018-01-13
implement wallet handling

jellegerbrandy:
not really a finished PR yet, but as i'm new to react I'd like to know if I am doing things the wrong way ..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           - v1-dependencies-{{ checksum "yarn.lock" }}
           - v1-dependencies-
       - run: bash <(curl https://get.parity.io -Lk)
-      - run: yarn run install
+      - run: yarn install
       - save_cache:
           paths:
             - node_modules
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,13 +16,13 @@ jobs:
           paths:
             - node_modules
           key: v1-dependencies-{{ checksum "yarn.lock" }}
-      - run: yarn run quality
-      - run: yarn run unit-test
+      - run: yarn quality
+      - run: yarn unit-test
       - run:
-          command: yarn run dev
+          command: yarn dev
           background: true
-      - run: yarn run chimp-test
-      - run: yarn run build
+      - run: yarn chimp-test
+      - run: yarn build
       - deploy:
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,18 +11,18 @@ jobs:
           - v1-dependencies-{{ checksum "yarn.lock" }}
           - v1-dependencies-
       - run: bash <(curl https://get.parity.io -Lk)
-      - run: yarn install
+      - run: yarn run install
       - save_cache:
           paths:
             - node_modules
           key: v1-dependencies-{{ checksum "yarn.lock" }}
-      - run: yarn quality
-      - run: yarn unit-test
+      - run: yarn run quality
+      - run: yarn run unit-test
       - run:
-          command: yarn dev
+          command: yarn run dev
           background: true
-      - run: yarn chimp-test
-      - run: yarn build
+      - run: yarn run chimp-test
+      - run: yarn run build
       - deploy:
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "build": "yarn run clean && yarn run prep && NODE_ENV=production webpack --progress --colors",
     "clean": "rm -rf build/",
-    "chimp-test": "chimp chimp.config.js --timeout 10000",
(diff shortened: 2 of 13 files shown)


src/scripts/utils/ParatiiLib.js
```diff
@@ -29,3 +29,28 @@ if (process.env.NODE_ENV === 'test') {
 }
 
 export const paratii: ParatiiLib = window.paratii
+
+export async function setupKeystore () {
+  let defaultPassword = 'password'
```
bent0b0x: I would put a lot of these strings in a new `constants/ParatiiConstants.js` file

bent0b0x:
Something I think we will want to do is first and foremost store all of this data (mnemonic, keystore name, etc) in our redux state. This way we can store it in immutable data structures and have it participate in our application state as will every other piece of meaningful application data. 

Then, to persist to local storage, I propose we look into [redux-localstorage](https://github.com/elgerlambert/redux-localstorage) (which admittedly I have never used, but basically allows you to seamlessly keep part of your redux store synced to `localStorage`. 

How does this sound?

src/scripts/utils/ParatiiLib.js
```diff
@@ -29,3 +29,28 @@ if (process.env.NODE_ENV === 'test') {
...
+      existingWalletIsValid = false
+    }
+  }
+  if (!existingWallet || existingWalletIsValid === false) {
+    console.log('Creating a new wallet')
+    mnemonic = await paratii.eth.wallet.newMnemonic()
+    localStorage.setItem(walletKey, JSON.stringify(paratii.eth.wallet.encrypt(defaultPassword)))
+    localStorage.setItem(mnemonicKey, mnemonic)
+  }
+  // console.log(`Your account is: ${}`)
```
bent0b0x: If we go with an approach like the one I propose in [this comment](https://github.com/Paratii-Video/paratii-portal/pull/53#issuecomment-357456716), then this logic would instead go into a redux thunk action in a `WalletActions.js` module

jellegerbrandy: It does not sound wrong :-) - but I am not sure I am the right person to judge :-)

Some of this state (e.g. the euser's ethereum address) is kept in the `paratii` object as well. Most of this state will probably be set just once in the lifetime of the app.  So we'd have 3 places to store the address:
   - paratii.config.address
   - redux
   - localstorage

jellegerbrandy: Actually the situation is mre complex. In localStorage, we store only the encrypted version of the wallet, not the wallet itself. (And similarly for the mnemonic: it is now stored in celar tes, but should be encrypted as well)

At this point, it is not clear to me what we would gain by keeping these values in te redux state.

bent0b0x: My understanding is that whether or not we have some of these data (e.g. address, encrypted wallet) and what their values are will control at least some logic in our application, yes? If so, then I think we definitely need it in the redux store. With redux and its connected container components, changes to the store/state will automatically re-render components based on store changes, if these components' properties coming from the store have changed. If we have data living outside of our store (e.g. in `localStorage`) then it is up to use to manually respond to these changes and re-render when necessary. That would introduce an additional layer of complexity to our app, I think, and we should avoid it.

One of the main tenants of Redux is single-source of truth/unidirectional data flow, and if we go against that pattern I think we could end up w/ a mess on our hands

.circleci/config.yml
```diff
@@ -16,13 +16,13 @@ jobs:
...
+      - run: yarn quality
+      - run: yarn unit-test
       - run:
-          command: yarn run dev
+          command: yarn dev
           background: true
-      - run: yarn run chimp-test
-      - run: yarn run build
+      - run: yarn chimp-test
+      - run: yarn build
```
bent0b0x: @jellegerbrandy I like to keep the `run` here and in all of our commands. Reason being if `yarn` ever has a built-in command with the same name as one of ours in `package.json`, then the `yarn` built-in will run instead of ours. We had this problem with `yarn clean` a week or two ago; safest to always include `run` for that reason (unless removing `run` adds some benefit I'm not aware of)

package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "build": "yarn run clean && yarn run prep && NODE_ENV=production webpack --progress --colors",
     "clean": "rm -rf build/",
-    "chimp-test": "chimp chimp.config.js",
+    "chimp-test": "chimp chimp.config.js --timeout 10000",
```
bent0b0x: Can you instead add this to `chimp.config.js`? I'd like to keep all of our config there instead of using the cli flags; just easier to view/edit later

.circleci/config.yml
jellegerbrandy: good point, will change back

package.json
jellegerbrandy: Yes, I will, also because chimp does not pick this up :-)

src/scripts/utils/ParatiiLib.js
jellegerbrandy: Let's discuss this by voice :-)

Issue #55, opened 2018-01-16
Systems Architecture (discussion)


jellegerbrandy:
# System Architecture

We have two consumer-facing apps:
 - portal 
 - player
 
This may in the future be extended with other things, like a mobile application.

Both make use fo paratii-mediaplayer (for playing files from ipfs) and paratii-lib (does the heavy lifting)

We have a number of requirements for the system setup:
 - The setup should avoid duplication of code as much as possible. 
  - Portal and player will share a certain amount of logic (how to save wallet info in the browser, how to create an account)
  - Portal and player share graphical design
 - We'd like both apps to share the same architecture stack (easier for developers and devops)
 - Both apps need a server side process to set the correct headers for our pages (it will also help with optimizing)
 - Incidentally, both apps will be deployed to the same domain and on the same machine something like `paratii.video/play/1` for the portal and `paratii.video/embed/play/1' for the embeddde version - so it makes sense to have them even share the same server-side process.

Proposal:

  -react/redux app
  - Add an express layer to add server-side rendering (using this  type of pattern https://css-tricks.com/server-side-react-rendering/ ) - i.e. adding express as an extra "indepdendent" laye ron top of the react app)

Open question: 
  - To me it seems logical to develop the two apps in the same repo, but @bent0b0X does not seem to agree... will it be easer to develop the two apps in two different repo's, or in a single one?  I cannot seem to find a good example of building different apps from a single code base.

bent0b0x:
After further consideration, I agree with your proposal to add this server layer in the same repo

bent0b0x:
FYI, I am working on this now and should have a PR up for review tonight (tomorrow morning for you guys ;P)

jellegerbrandy:
With regard to the PR, here's what I had in mind: 

 * With regard to "system architecture", there is really no distinction between "embeddable player" and "portal". There is a single codebase that creates components that can be embedded, and creates pages for the portal. 
 * The system needs a server-side process that sets the correct headers on the pages (and serves the right content at the appropriate URLs). This can be a relative simple express-layer on top of the react app.
 * For optimizazation purposes (i.e. this is more a devops question), it makes ssense to have different webpack builds for different parts of the app: in particular, the pages that are to be embedded should be as light-weight as possible.

bent0b0x:
I think the PR is good on points 1 and 3 (but feel free to refute that :) ). On the second point, yes right now the server process is only for serving up the embed html views. I can rework the server to encompass the entire app

[...]

PR #57, opened 2018-01-18
WIP do not merge - Add video embed server

bent0b0x:
**Work Done**
- add lightweight express server to portal with the purpose of serving up embedded video content

**Demo**

embeddemo

- [ ] fix tests
- [ ] update readme


package.json
```diff
@@ -11,18 +11,20 @@
     "chimp-test": "chimp chimp.config.js",
     "chimp-test:watch": "yarn run chimp-test --watch",
     "dev": "concurrently \"yarn run dev-client\" \"yarn run dev-server\" \"yarn run dev-parity\"",
-    "dev-parity": "yarn parity",
-    "dev-client": "yarn run clean && yarn run prep && NODE_ENV=test webpack --watch --colors --progress",
-    "dev-server": "NODE_ENV=test nodemon --watch src/server --watch node_modules src/server/index.js localhost:8080 ",
+    "dev-parity": "yarn run parity",
```
bent0b0x: Just as a note, I think we should always include `run` because if we unknowingly name a command the same as the name of a built-in `yarn` command, leaving out `run` will cause the built-in command to be run instead (we hit this issue w/ `yarn clean` before)

package.json
```diff
@@ -11,18 +11,20 @@
     "chimp-test": "chimp chimp.config.js",
     "chimp-test:watch": "yarn run chimp-test --watch",
     "dev": "concurrently \"yarn run dev-client\" \"yarn run dev-server\" \"yarn run dev-parity\"",
-    "dev-parity": "yarn parity",
-    "dev-client": "yarn run clean && yarn run prep && NODE_ENV=test webpack --watch --colors --progress",
-    "dev-server": "NODE_ENV=test nodemon --watch src/server --watch node_modules src/server/index.js localhost:8080 ",
```
bent0b0x: I'm not sure why this was added in, but I don't think we want `NODE_ENV=test` unless we are actually running tests

package.json
```diff
@@ -8,21 +8,23 @@
   "scripts": {
     "build": "yarn run clean && yarn run prep && NODE_ENV=production webpack --progress --colors",
     "clean": "rm -rf build/",
-    "chimp-test": "chimp chimp.config.js",
-    "chimp-test:watch": "yarn run chimp-test --watch",
+    "chimp-test": "concurrently --kill-others --success first \"NODE_ENV=test yarn run dev\" \"chimp chimp.config.js\"",
```
bent0b0x: Changing our test commands so that we don't set `NODE_ENV=test` every time we build just for dev. Now we only set it when truly in test mode

bent0b0x:
@jellegerbrandy interested to hear what you think of this setup.

Here is what we have in this PR:

- Now, `webpack` will produce multiple outputs. This is all managed via the same build command
  - portal js bundle
  - embed js bundle
  - embed stylesheet

- There is a new command (`$ yarn run start-embed`) that will boot up the express server and by default listen on port 3000. This will serve up an html page that creates a `paratii-mediaplayer` instance based on the url params

What do you think?

webpack.config.js
```diff
@@ -1,31 +1,36 @@
...
+const config = {
+  entry: {
+    bundle: prod
+      ? [scriptsDir + "/index.js"]
+      : [
+          "react-hot-loader/patch",
+          scriptsDir + "/index.js",
+          "webpack-hot-middleware/client?quiet=true"
+        ],
+    'embed/bundle': embedDir + "/client/index.js"
```
bent0b0x: This is how we can specify multiple entry points that will be bundled into separate outputs

webpack.config.js
```diff
@@ -1,31 +1,36 @@
...
+  },
   output: {
     path: buildDir,
     publicPath: "/",
-    filename: "bundle.js"
+    filename: "[name].js"
```
bent0b0x: And this allows the bundles to be named dynamically based on the config a few lines above

package.json
bent0b0x: In addition, `yarn run chimp-test` will now boot up `parity`, the dev server, and anything else we need. So you don't have to worry about running multiple commands.

bent0b0x: Also, want to be explicit about the `--kill others --success first` params passed to `concurrently` as they are not things I had used before.

Basically, by default `concurrently` will continue running as long as any of its child processes have not exited. In our case, `yarn run dev` never exits (unless the bundle fails), so even after our chimp tests complete the `concurrently` process would never exit and our build would not pass. 

So, the `--kill-others` flag partially solves that problem by killing all processes once one finishes (in our case the `chimp` tests). Secondly, the `--success first` tells `concurrently` to exit with a success or failure based on the success/failure of the first process to exist. If we don't use this, `concurrently` will always exit w/ a non-zero code.

src/scripts/embed/server/views/video.handlebars
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
```
bent0b0x: This is the view template for our embedded video; please let me know if anything looks wrong/missing/etc.

src/scripts/embed/server/views/video.handlebars
```diff
@@ -0,0 +1,23 @@
...
+<html>
+<head>
+  <title>{{title}}</title>
+  <link rel="stylesheet" href="/embed.css">
+  <script src="/bundle.js"></script>
```
bent0b0x: This `bundle.js` is not the same one used by the portal. Rather, it is `build/embed/bundle.js`, which is only concerned w/ dependencies required to render our embeddable player.

package.json
eliawk: this is super nice, even if i know @jellegerbrandy doesn't like concurrently, because i want to have all the different processes run in multiple terminal windows!

jellegerbrandy: yes, exactly. Nothing against using concurrently, but I'd also like a command that just calls chimp without starting other processes (which I can start by hand, i.e. the original chimp-test command)

jellegerbrandy:
cf. my comment on #55

package.json
jellegerbrandy: ah, also, it messes up completely the circleci ouptut doing this stuff concurrently. It is useful there to start the processes explicitly separately, to debug the problems more easily.

bent0b0x: I will revert that change before merging 👍

bent0b0x:
Quick update as to the status of this PR:

- The express server (that was previously just the dev server) is being promoted to actually serve up all of our content, in production
  - this includes the portal app
  - and also embedded content
- the embedded content is going to have its own js bundle and css bundle, which should be much smaller than the portal bundle(s). It will share some code w/ the portal (the `Play` container for now)
- This PR also contains some refactors to how we are using `paratii-lib`, including creating an api to initialize the lib and then provide access to a singleton instance of the lib. With this change, it is the responsibility of the app(s)/tests init `paratii-lib` before using it (which if not in `production` mode will deploy the contracts for you)

src/scripts/actions/VideoActions.js
```diff
@@ -11,7 +11,7 @@ export const videoDataLoaded = createAction(VIDEO_DATA_LOADED)
 
 export const fetchVideo = (id: string) => async (dispatch: Dispatch<*>) => {
   console.log('fetching video info for video with id: ' + id)
-  let videoInfo = await paratii.eth.vids.get(id)
+  let videoInfo = await paratii().eth.vids.get(id)
```
bent0b0x: I am proposing this to be the new model for accessing the paratii-lib instance. Simply call `paratii()` to get the singleton, and only after the surrounding app has called `initParatiiLib` asynchronously

src/scripts/embed/index.js
```diff
@@ -0,0 +1,43 @@
+/* @flow */
+
```
bent0b0x: This is the separate embed FE app gets loaded into the browser on requests to `/embed/:id` routes.

src/scripts/index.js
```diff
@@ -5,43 +5,36 @@ import ReactDOM from 'react-dom'
...
-
-ReactDOM.render(
-  <Provider store={store}>
-    <AppContainer>
-      <Root />
-    </AppContainer>
-  </Provider>,
-  root
-)
+initParatiiLib()
```
bent0b0x: Here is an example of initializing paratii-lib before rendering the app

src/server/routes/embed.js
```diff
@@ -0,0 +1,29 @@
...
+  const thumbnailUrl = 'http://paratii.video/imagens/cropped-logo_colorido_horizontal.png'
+  const url = `https://portal.paratii.video/play/${id}`
+
+  res.send(`
+    <!DOCTYPE html>
+    <html>
+      <head>
+        <title>${title}</title>
+        <link rel="stylesheet" type="text/css" href="/embed/index.css">
+        <meta name="og:title" content="${title}"/>
```
bent0b0x: Here is where we set our `html` headers. The rest of the embed content is rendered on the FE

src/scripts/index.js
bent0b0x: And I may want to think of another approach to this, as this noticeably slows down the rendering on the FE.

bent0b0x:
@jellegerbrandy feel free to take another look at this. The major change now is as follows:

`paratii-lib` is now instantiated with a config that (at least partially) comes from a config file. Which config file is dependent on `process.env.NODE_ENV` at build-time. In dev mode, we run `deployContracts` before booting up the app and write the `registryAddress` to `config/development.json`, which the app then accesses.

I think we are close to being able to merge this. My major hesitation right now is the fact that the app cannot run right now due to `web3.js` package issues that prevent me from performing further testing; hope to have those resolved (with the help of @geckoslair 😄 ) soon!

jellegerbrandy:
Yes, I think i tis pretty much time to merge this, once the tests are running on circle

config/development.json
```diff
@@ -0,0 +1,6 @@
+{
+  "provider": "http://localhost:8545/rpc",
+  "address": "0x9e2d04eef5b16CFfB4328Ddd027B55736407B275",
+  "privateKey": "399b141d0cc2b863b2f514ffe53edc6afc9416d5899da4d9bd2350074c38f1c6",
+  "registryAddress": "0xB082D0cB94aefB147E7910CA579209f47c694C91"
```
jellegerbrandy: IF I understand correctly, this is the configuration used by the developer when playing wiht the app locally against a test chain. in that case, typically no contract will be deployed - and so we do not have the registyr address available. And so this line shoud be removed

PR #59, opened 2018-01-18
8 uploader (don't merge yet)

jellegerbrandy:


src/scripts/components/VideoForm.js
```diff
@@ -0,0 +1,44 @@
...
+`
+
+class VideoForm extends Component<Props, void> {
+  render () {
+    const { onSubmit, onInputChange } = this.props
+    return (
+      // <UploadProgressBarContainer />
+      <Form onSubmit={onSubmit}>
+        <Input
+          id='video-title'
```
bent0b0x: Is this id (and that of the other elements in this component) actually needed?

src/scripts/components/UploadFile.js
```diff
@@ -1,24 +1,33 @@
...
+  constructor (props) {
+    super(props)
+    this.state = {}
+    this.onFileChosen = this.onFileChosen.bind(this)
+  }
+
+  onFileChosen (e) {
+    const file = e.target.files[0]
+    this.props.onFileChosen(file)
+    this.setState({file: e.target.files[0]})
```
bent0b0x: I don't think we should need to store `file` in `state`. Ideally, IMO, `onFileChosen` would set the file in the redux store and then the container would pass down `file` as a prop

src/scripts/actions/UploadActions.js
```diff
@@ -0,0 +1,65 @@
...
+
+export const saveVideoInfo = (videoInfo: Object) => (dispatch: Dispatch<*>) => {
+  console.log('Saving video info')
+  // TODO: paratii-lib shoudl generate a fresh id, we now use a placeholder
+  videoInfo.id = `foo_${Math.floor(Math.random() * 100000)}`
+  // the owner is the user that is logged in
+  videoInfo.owner = paratii.config.account.address
+  // TODO: paratii-lib has no support for the description yet (there is an issue)
+  // TODO: once that support is there, the following line should be deleted
+  delete videoInfo.description
```
bent0b0x: Why this deletion?

src/scripts/components/VideoStatus.js
```diff
@@ -0,0 +1,47 @@
+/* @flow */
+
+import React, { Component } from 'react'
+// import styled from 'styled-components'
+// import Button from './foundations/Button'
+// import Input from './foundations/Input'
+import Wrapper from './foundations/Wrapper'
+
+type Props = {
+  video: Object,
```
bent0b0x: Can this instead be a `Video` record? That is what the redux store would contain anyways (and it looks like that is the more accurate type in this case already)

src/scripts/components/VideoStatus.js
```diff
@@ -0,0 +1,47 @@
...
+  videoFromPTI: Object
+};
+
+class VideoStatus extends Component<Props, void> {
+  render () {
+    let video = this.props.video
```
bent0b0x: I think we should prefer `const` over `let` whenever a variable is never re-assigned

src/scripts/components/VideoStatus.js
```diff
@@ -0,0 +1,47 @@
...
+    return (
+      <Wrapper>
+        <pre id="video-status">
+            Your Video is now being uploaded/beging transcoded/all ready for sharing!
+          <br />
+            Once it is ready,
+            You can share your video here [link] or embed it like this [link]
```
bent0b0x: Probably something we want to handle later, but we'll need to come up with a plan for internationalization

src/scripts/components/VideoStatus.js
```diff
@@ -0,0 +1,47 @@
...
+
+import React, { Component } from 'react'
+// import styled from 'styled-components'
+// import Button from './foundations/Button'
+// import Input from './foundations/Input'
+import Wrapper from './foundations/Wrapper'
+
+type Props = {
+  video: Object,
+  videoFromPTI: Object
```
bent0b0x: What is this prop, exactly?

src/scripts/containers/UploadFileContainer.js
```diff
@@ -1,9 +1,50 @@
...
+
+const mapStateToProps = (state: RootState) => ({})
 
-const mapDispatchToProps = () => ({})
+const mapDispatchToProps = dispatch => ({
+  upload: bindActionCreators(upload, dispatch)
+})
 
-export default connect(mapStateToProps, mapDispatchToProps)(UploadFile)
+export default connect(mapStateToProps, mapDispatchToProps)(UploadFileContainer)
```
bent0b0x: This isn't a huge point, but thus far we have been separating the `containers` from the `components`. In effect, this would mean moving the component class to a `components/UploadFile.js` module and importing it into this container. Then this container would just define the `mapState/DispatchToProps` and `connect` call.

One of the benefits of doing this is so that it is easier to unit test the components (but that's not really a huge deal). A more important point is that we should just decide which way to do it, because as it stands we are using two different techniques in the same repo.

src/scripts/reducers/UploadReducer.js
```diff
@@ -0,0 +1,46 @@
...
+      isUploading: true
+    })
+  },
+  [UPLOAD_PROGRESS]: (
+    state: UploadRecord,
+    { payload }: Action<{progress: number}>
+  ): UploadRecord => {
+    return state.mergeDeep({
+      progress: { value: payload.progress }
+    })
```
bent0b0x: An alternative could be `state.setIn(['progress', 'value'], payload.progress)`

PR #84, opened 2018-01-26
fixing parity version

geckoslair:


jellegerbrandy:
Hey @geckoslair , your upgarde of web3js broke the patch (which, I presume, isnot needed anymore). Can you remove the patch?

jellegerbrandy:
Okay, added this stuff by hand trying to fix dev.
Thanks!
Anyway, the problem is, of course, not the parity version; the problem is the configuration file :-)

PR #89, opened 2018-01-29
81 - show avatar

eliawk:
add react-blockies, the avatar is now based on the eth address.

`react-blockies` is a simple port of [](https://github.com/alexvandesande/blockies to a react component.)


eliawk:
@bent0b0x i'm fighting with `stylelint`. it auto change a file and then it tells me there are errors which it has created!

bent0b0x:
@eliawk I don't think `stylelint` changes any files. Are you sure it is not `prettier-standard` doing the changes?

bent0b0x:
If [this rule](https://stylelint.io/user-guide/rules/declaration-colon-newline-after/) conflicts w/ prettier I would disable it.

jellegerbrandy:
fixed the linter problem with #90,

PR #93, opened 2018-01-30
WIP DO NOT MERGE 8 uploader

bent0b0x:


src/scripts/components/VideoList.js
```diff
@@ -0,0 +1,41 @@
...
+`
+
+class VideoList extends Component<Props, void> {
+  render () {
+    const { uploads, onItemClick } = this.props
+    return (
+      <Wrapper>
+        {uploads
+          .entrySeq()
+          .map(([k, v]) => (
```
bent0b0x: need to name these variables. I have no idea what `k` and `v` are :)

src/scripts/components/VideoList.js
```diff
@@ -0,0 +1,41 @@
...
+            <VideoListItem key={k} id={k} item={v} onClick={onItemClick} />
```
bent0b0x: If `k` is the video id, then this is fine. If it is an array/list index, it cannot be used as a react `key`.

src/scripts/components/VideoListItem.js
```diff
@@ -0,0 +1,57 @@
...
+    props.theme.fonts.family ? props.theme.fonts.family : 'Monospace'},
+    sans-serif;
+  display: flex;
+  flex-direction: column;
+  border: 1px solid grey;
+  padding: 10px;
+`
+
+const Label = styled.p`
+  color: white;
```
bent0b0x: Please use the colors @jrgarou defined. Example [here](https://github.com/Paratii-Video/paratii-portal/blob/3f1abb724d25836dc49858869f226e2e6cd4c951/src/scripts/components/structures/Main.js#L9), and you may need/want to edit the [themes file](https://github.com/Paratii-Video/paratii-portal/blob/3f1abb724d25836dc49858869f226e2e6cd4c951/src/scripts/components/foundations/base/Themes.js)

src/scripts/components/VideoListItem.js
```diff
@@ -0,0 +1,57 @@
...
+
+  handleClick () {
+    this.props.onClick(this.props.id)
+  }
+
+  render () {
+    const item = this.props.item
+    console.log(this.props)
+    let status = ''
+    if (item.getIn(['uploadStatus', 'name']) === 'running') {
```
bent0b0x: Please abstract this to an instance method.

Actually, this should be in a `createSelector` selector and passed in via a prop. Essentially all calculated pieces of state that have to do w/ our application data (as opposed to state of a UI element, e.g. dropdown open vs. closed) should be in a selector

src/scripts/containers/VideoFormContainer.js
```diff
@@ -0,0 +1,72 @@
...
+import type { RootState } from 'types/ApplicationTypes'
+
+type Props = {
+  updateVideoInfo: (id: string, title: string, description: string) => void,
+  videoInfo: VideoInfoRecord
+}
+
+class VideoFormContainer extends Component<Props, void> {
+  constructor (props) {
+    console.log('constructor')
```
bent0b0x: delete logging statement

src/scripts/containers/VideoFormContainer.js
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  handleInputChange (input, e) {
+    this.setState({
+      [input]: e.target.value
+    })
+  }
+
+  handleSubmit (e) {
+    console.log('handleSubmit', e)
```
bent0b0x: logging

src/scripts/containers/VideoManagerContainer.js
```diff
@@ -0,0 +1,72 @@
...
+
+  render () {
+    const { uploads } = this.props
+    const { selectedUpload } = this.state
+    return (
+      <Wrapper>
+        {this.props.uploads.count() > 0 && (
+          <VideoList onItemClick={this.onVideoListItemClicked} />
+        )}
+        {selectedUpload === null && <UploadFile />}
```
bent0b0x: Can use a ternary so the logic doesn't have to be written twice:

```
selectedUpload === null
? <UploadFile />
: <VideoForm /> //etc
``

File: src/scripts/reducers/UploadReducer.js [hunk 8]
```diff
@@ -74,4 +86,4 @@ const reducer = {
   }
 }
 
-export default handleActions(reducer, fromJS(new UploadRecord()))
+export default handleActions(reducer, Map({}))
```
bent0b0x: Why is this no longer an `UploadRecord`?

src/scripts/selectors/UploadSelectors.js
```diff
@@ -1,48 +1,48 @@
-import { createSelector } from 'reselect'
+// import { createSelector } from 'reselect'
 
-import { getUpload } from 'selectors/index'
+// import { getUpload } from 'selectors/index'
 
 /* Upload */
-export const getUploadProgress = createSelector(
```
bent0b0x: Why delete all of these?

src/scripts/containers/VideoManagerContainer.js
```diff
@@ -0,0 +1,72 @@
...
+class VideoManagerContainer extends Component<Props, void> {
+  constructor (props) {
+    super(props)
+    this.state = { selectedUpload: null }
+    this.onVideoListItemClicked = this.onVideoListItemClicked.bind(this)
+  }
+
+  onVideoListItemClicked (id: string) {
+    console.log('onVideoListItemClicked')
+    this.setState({
```
bent0b0x: I would change this to be dispatching an action to update an `editingVideoId` in the redux store

src/scripts/actions/VideoActions.js
```diff
@@ -17,3 +17,8 @@ export const fetchVideo = (id: string) => async (dispatch: Dispatch<*>) => {
   console.log('Received video info')
   dispatch(videoDataLoaded(videoInfo))
 }
+
+export const setCurrentVideo = (id: string) => (dispatch: Dispatch<*>) => {
```
bent0b0x: There is no need to have `setCurrentVideo` right now, all it does is call `dispatch(videoDataLoaded())`. You can instead pass `videoDataLoaded` to the container

src/scripts/containers/VideoManagerContainer.js
```diff
@@ -1,16 +1,21 @@
+import { Map } from 'immutable'
```
bent0b0x: FYI in the rest of the repo we have been keeping the `components/` files separate from the `containers/` files, rather than merging them into one. We can go either way, but we should try to be consistent.

Keeping them separate does make it a bit easier to _unit_ test, but that's not something we're doing extensively right now.

src/scripts/types/ApplicationTypes.js
```diff
@@ -30,7 +30,7 @@ export type Action<T> = {
 }
 
 export type RootState = {
-  video: ?VideoRecord,
+  selectedVideo: ?VideoRecord,
```
bent0b0x: We might want to refine our state a bit more. What does `selectedVideo` mean, exactly? Is this going to be separate to a video you select w/ the purposes of just playing the video? Or is this only a video you are editing

src/scripts/reducers/UploadReducer.js
jellegerbrandy: because it is a mapping of ids to Uploadrecords.
So this line looks good to me.
What looks fishy to me is the fact that all functions in the reducers are declared to operate on UploadREcords ( `state: UploadRecord`) but in fact do not

bent0b0x: Yes, this reducer looks pretty much screwed up.

I think we should keep `UploadRecord` and it should itself have a prop of `uploadedVideos` (or even just `videos`) that is this map of ids to `VideoRecord`

src/scripts/containers/VideoManagerContainer.js
jellegerbrandy: yes. Actualy, this state was still a left-over, removed it

PR #95, opened 2018-01-31
Uploader ui

jrgarou:
On this pull request:
- **FilesUploader components/widgets**
- **TextField(input text) components/foundations/forms**
- [Text sizes ](https://github.com/Paratii-Video/paratii-portal/blob/uploader-ui/src/scripts/components/foundations/base/Typography.js#L12-L16)
- [MainSvg](https://github.com/Paratii-Video/paratii-portal/blob/uploader-ui/src/scripts/components/foundations/svgs/MainSvg.js) with all reusable svgs

## FilesUploader components/widgets:
```

import FilesUploader from '....components/widgets/FilesUploader'

<FilesUploader 
    onFileChosen={this.handleFileChosen}
    onUploadRequested={this.handleUploadRequested}
/>

```

image

## TextField(input text) components/foundations/forms:
```

import TextField from '...components/foundations/forms/TextField'


<TextField
    label="Or upload from Youtube or Vimeo"
    helper="i.e.: http://youtube.com/videoID or http://vimeo.com/videoID"
    error={false}
/>


```
image


src/scripts/components/structures/MainHeader.js
```diff
@@ -4,17 +4,14 @@ import MainHeaderLogo from '../widgets/MainHeaderLogo'
 import SearchInput from '../widgets/SearchInput'
 import MainNavigation from '../widgets/MainNavigation'
 import { Link } from 'react-router-dom'
-import Blockies from 'react-blockies'
 
 type Props = {
   children: Object
 }
```
bent0b0x: We have two files named `MainHeader.js` that live in different subdirectories. Can we change one of them? I would think one of them must be named incorrectly

src/scripts/components/foundations/forms/TextField.js
```diff
@@ -0,0 +1,116 @@
...
+      value: ''
+    }
+
+    this.handleChange = this.handleChange.bind(this)
+    this.handleKeyUp = this.handleKeyUp.bind(this)
+    this.handleClasses = this.handleClasses.bind(this)
+  }
+
+  handleClasses () {
+    let klass = this.state.filled ? 'filled ' : ''
```
bent0b0x: Curious, why are we using class names here instead of styled components? Just want to understand.

Also, for this kind of class name logic:  we should use the [classnames](https://github.com/JedWatson/classnames) package.

It basically let's you write code like this:

```javascript
const className = classnames(this.props.className, {
  filled: this.state.filled,
  error: this.props.error
});
```

This creates a class string where `this.props.className` is always present, and `filled` and `error` are added conditionally.

src/scripts/components/structures/MainHeader.js
jrgarou: Whatever you want.
What is this `/components/MainHeader.js`?

src/scripts/components/foundations/forms/TextField.js
jrgarou: Nice to know, @bent0b0x.
This is states of the component.

Let me try this

src/scripts/components/App.js
```diff
@@ -44,7 +43,6 @@ class App extends Component<Props, void> {
         <MainTemplate>
           <MainHeader />
           <Main>
-            <Route exact path="/" component={Home} />
```
eliawk: Why did you remove the home page?

src/scripts/components/foundations/svgs/FilesUploaderSvg.js
```diff
@@ -0,0 +1,42 @@
...
+
+type Props = {}
+
+const SVG = styled.svg`
+  display: block;
+  height: 100%;
+  width: 100%;
+`
+
+class FilesUploaderSvg extends Component<Props, void> {
```
eliawk: having a components for just a static svg it seems quite too much.

src/scripts/components/structures/MainHeader.js
```diff
@@ -93,10 +90,9 @@ class MainHeader extends Component<Props, void> {
                 </ButtonBucketSVG>
               </ButtonBucket>
               <ProfileAvatarLink to="/signup">
-                <Blockies
```
eliawk: hey you remove blokies instead of this static image!!

src/scripts/components/structures/MainHeader.js
```diff
@@ -4,17 +4,14 @@ import MainHeaderLogo from '../widgets/MainHeaderLogo'
 import SearchInput from '../widgets/SearchInput'
 import MainNavigation from '../widgets/MainNavigation'
 import { Link } from 'react-router-dom'
-import Blockies from 'react-blockies'
```
eliawk: we need this!

eliawk: we need this!

src/scripts/components/foundations/svgs/FilesUploaderSvg.js
bent0b0x: I think we should come up w/ a universal svg solution. This may be a fine way to go, we can even make this into a stateless functional component so it is cleaner. But I think we definitely can't revert back to the `Svgs.js` component that this PR deletes. That component would render all of our app SVGs every time we use one svg, which is not scalable. This method to me is preferable.

eliawk: after reading this https://css-tricks.com/creating-svg-icon-system-react/ i agree with you guys.
we could pass props to svg and make awesome stuff.

PR #100, opened 2018-02-01
WIP- 8 uploader

eliawk:
there is still a flow error:

```
Launching Flow server for /Users/elia/Siti/paratii-portal
Spawned flow server (pid=6489)
Logs will go to /private/tmp/flow/zSUserszSeliazSSitizSparatii-portal.log
Monitor logs will go to /private/tmp/flow/zSUserszSeliazSSitizSparatii-portal.monitor_log
Error: src/scripts/reducers/VideoReducer.js:17
 17:     console.log('Setting selected video to:' + payload.id)
                                                            ^^ property `id`. Property not found in
 17:     console.log('Setting selected video to:' + payload.id)
                                                    ^^^^^^^ String


Found 1 error
```


src/scripts/components/UploadFile.js
```diff
@@ -1,38 +0,0 @@
...
-              Try dropping some files here, or click to select files to upload.
-            </p>
-          </Dropzone>
-        </section>
-      </Wrapper>
-    )
-  }
-}
-
-export default UploadFile
```
eliawk: this file was a duplicate and not used anymore

bent0b0x:
@eliawk  re: the `flow` error. It would appear that `payload` is typed to be a `String`. Is there an `Action<string>` typdef anywhere? 

Also, we should just remove the `console.log`, we have too many in the app :)

src/scripts/components/Debug.js
```diff
@@ -3,20 +3,19 @@ import styled from 'styled-components'
 import { paratii } from '../utils/ParatiiLib'
 import type { RootState } from 'types/ApplicationTypes'
 
-const DebugBox = styled.header`
-  background-color: #fff;
-  border: 2px dashed red;
+const DebugBox = styled.div`
+  background-color: #ccc;
+  border: 4px dashed #f55;
```
bent0b0x: I think we should avoid any hex colors directly in components and instead, if we have new colors, define them in our colors module

src/scripts/containers/VideoManagerContainer.js
```diff
@@ -45,15 +45,25 @@ class VideoManagerContainer extends Component<Props, void> {
...
     return (
-      <Wrapper>
-        <VideoList onItemClick={this.onVideoListItemClicked} />
-        {selectedVideo === null && <UploadFile />}
-        {selectedVideo !== null && <VideoForm />}
+      <div>
+        <Wrapper>
+          {videolist}
+          {selectedVideo === null && <UploadFile />}
+          {selectedVideo !== null && <VideoForm />}
```
bent0b0x: I would rewrite this as a ternary so that we don't basically write the logic 2x:

```javascript
{ selectedVideo === null ? <UploadFile /> : <VideoForm /> }
```

src/scripts/components/Debug.js
eliawk: i agree but this is just a temporary ugly debug box, just for dev

src/scripts/containers/VideoManagerContainer.js
eliawk: agree, i'll change this

PR #101, opened 2018-02-02
94 state definition

jellegerbrandy:


src/scripts/actions/UploadActions.js
```diff
@@ -20,7 +20,7 @@ import {
 
 import type { Dispatch } from 'redux'
 
-import VideoInfoRecord from 'records/VideoInfoRecords'
+import VideoInfoRecord from 'records/VideoRecords'
```
bent0b0x: Can the import be renamed too? to `VideoRecord`

src/scripts/components/Play.js
```diff
@@ -8,12 +8,13 @@ import VideoRecord from 'records/VideoRecords'
 import VideoOverlay from 'components/VideoOverlay'
 
 import type { RouteMatch } from 'types/ApplicationTypes'
+import Debug from 'containers/DebugContainer'
```
bent0b0x: Should this be removed before merging?

src/scripts/components/Play.js
```diff
@@ -53,7 +54,8 @@ const Player = styled.div`
 class Play extends Component<Props, void> {
   componentDidMount (): void {
     const videoId = this.props.match.params.id
-
+    console.log('componentDidMount')
```
bent0b0x: Remove these logging statements before merging?

src/scripts/components/VideoForm.js
```diff
@@ -26,14 +25,20 @@ class VideoForm extends Component<Props, Object> {
...
   }
 
+  componentWillReceiveProps (nextProps: Props): void {
+    this.setState(nextProps.selectedVideo)
+    this.setState({
+      id: nextProps.selectedVideo.id,
+      title: nextProps.selectedVideo.title,
+      description: nextProps.selectedVideo.description
+    })
+  }
```
bent0b0x: This should not be necessary. If we already have all of the data in `props`, there is no need to store it in state. We should just rely on grabbing this information from `props`.

bent0b0x: Also, two `setState` operations in a row doesn't seem necessary, it could be combined into one (but I still recommend zero here)

src/scripts/components/VideoForm.js
```diff
@@ -52,13 +57,15 @@ class VideoForm extends Component<Props, Object> {
...
+    console.log(this.state.id)
     return (
       <Form>
-        Editing video with id: {this.props.selectedVideo.id}
+        Editing video {this.state.id}
         <Input
           id="video-id"
           type="hidden"
-          value={this.props.selectedVideo.id}
+          value={this.state.id}
```
bent0b0x: What is the problem w/ just using `props`?

src/scripts/components/VideoOverlay.js
```diff
@@ -30,17 +30,28 @@ const Overlay = styled.div`
 
 class VideoOverlay extends Component<Props, void> {
   render () {
```
bent0b0x: FYI on my branch I'm re-working this component to match the designs so whatever you have here will likely change

src/scripts/containers/UploadVideoStatusContainer.js
```diff
@@ -4,11 +4,10 @@ import React, { Component } from 'react'
 import VideoStatus from 'components/VideoStatus'
 import type { RootState } from 'types/ApplicationTypes'
 import VideoRecord from 'records/VideoRecords'
-import UploadRecord from 'records/UploadRecords'
 
 type Props = {
-  video: VideoRecord,
-  state: UploadRecord
+  video: VideoRecord, // TODO: seems a bit too much
+  state: VideoRecord
```
bent0b0x: We definitely should have a `prop` called `state`, that is bound to confuse the hell out of people :P. What does this actually mean?

src/scripts/records/UploadRecords.js
```diff
@@ -1,20 +1,20 @@
-/* @flow */
```
bent0b0x: Uncomment this file?

src/scripts/records/VideoInfoRecords.js
```diff
@@ -1,18 +1,18 @@
...
-  owner: null
-}) {
-  id: ?string
-  title: ?string
-  description: ?string
-  ipfsHash: ?string
-  owner: ?string
-}
-export default VideoInfo
+// /* @flow */
```
bent0b0x: Why all commented?

src/scripts/records/VideoRecords.js
```diff
@@ -1,27 +1,36 @@
...
-  ipfsData: string;
-  ipfsHash: string;
-  owner: string;
-  price: string;
-  thumbnailUrl: string;
-  title: string;
-  url: string;
+  blockchainStatus: AsyncTaskStatusRecord
+  description: string
+  filename: string
```
bent0b0x: One thing to look at is will these properties always be set? Some of them may need to be optionals if `paratii-lib` will not always include these fields (e.g. `id: string` => `id?: string` or `id: ?string`). The former is if the key is sometimes left out, and the latter is if the key is always there but the value might be `undefined` or `null`. You could even do `id?: ?string` if both are possible.

src/scripts/reducers/SelectedVideoReducer.js
```diff
@@ -0,0 +1,35 @@
...
+} from 'constants/ActionConstants'
+// import VideoRecord from 'records/VideoRecords'
+import type { Action } from 'types/ApplicationTypes'
+
+const reducer = {
+  [VIDEO_SELECT]: (
+    state: string,
+    { payload }: Action<{ id: string }>
+  ): string => {
+    console.log('Setting selected video to:' + payload.id)
```
bent0b0x: Remove logging?

src/scripts/reducers/SelectedVideoReducer.js
```diff
@@ -0,0 +1,35 @@
...
+    return payload.id
+    // return new VideoRecord(payload)
+  }
+  // ,
+  // [VIDEO_DATA_PROGRESS]: (
```
bent0b0x: Can we remove all of these comments? My fear is that they will stay around forever...

src/scripts/selectors/index.js
```diff
@@ -1,12 +1,15 @@
...
 
 import VideoRecord from 'records/VideoRecords'
-import UploadRecord from 'records/UploadRecords'
 import UserRecord from 'records/UserRecords'
 import type { RootState } from 'types/ApplicationTypes'
 
 /* Videos */
-export const getVideo = (state: RootState): ?VideoRecord => state.selectedVideo
+export const getVideo = (state: RootState): ?VideoRecord => {
+  let videoId = state.selectedVideo
```
bent0b0x: nitpicky, but let's try to use `const` over `let` when the value will never change (maybe I can just turn on an `eslint` rule for this ;) )

src/scripts/components/VideoForm.js
jellegerbrandy: I'm not sure, how found the form ahndling both state and props.
can we manage the state of the form with the props?

src/scripts/components/VideoForm.js
jellegerbrandy: seems reasonable to me

src/scripts/reducers/index.js
```diff
@@ -2,12 +2,12 @@
...
+import selectedVideo from 'reducers/SelectedVideoReducer'
 import user from 'reducers/UserReducer'
-import uploads from 'reducers/UploadReducer'
+import videos from 'reducers/VideosReducer'
 
 export default combineReducers({
-  selectedVideo: video,
-  user,
-  uploads
+  selectedVideo: selectedVideo,
```
bent0b0x: Can this be changed to `selectedVideoId`? The way it is currently named, it is implied this this is a video record not a number/string

src/scripts/components/Play.js
jellegerbrandy: we are in full dev mode, and I find it very useful. We should remove it once we go public, obviously.

src/scripts/records/VideoInfoRecords.js
jellegerbrandy: We have one single VideoRecord now that replaces "UploadRecords" and "VideoInfoRecords". I'll remove the files..

src/scripts/components/VideoForm.js
bent0b0x: It depends on what we are trying to accomplish. You could handle the form values purely w/ props, purely w/ state, or a combination of both. 

This is _probably_ what we want to do:

- in the `constructor`, set our initial state (wrt the form values) based on `props`. 
- do not update state wrt form values in `componentWillReceiveProps`
- When form values change, update the respective values in `state`
- When submitting the form, trigger an action that persists the change and when successful updates the redux store w/ the new relevant values. 

How does that sound?

bent0b0x: This keeps our true values in `props`, and while we are editing the form since `props` is not changing we have our source of truth maintained until the user successfully submits the changes on the form. This allows us to easily cancel our changes if the user does not submit the form, for example.

PR #110, opened 2018-02-06
22 player design

bent0b0x:
**Issue**
[paratii-mediaplayer#22](https://github.com/Paratii-Video/paratii-mediaplayer/issues/22)

**Work Done**
- Implement basics of player design (although it is still very rough). This, though, sets some groundwork for building up and iterating on the design
- Establish a pattern for how we can split our code and load it dynamically. This pattern allows the player component to be shared by both the portal app and the embed app, and for functionality that differs between the two we can load the required code only in each version of the component that requires it

**Disclaimer**
Obviously, the video overlay has no transition. That will be coming next, so do not think that this (and anything else design-wise) in this PR is final 😬 

**Demo** (please ignore the long video load times, that will be tackled separately 😄 )

*Portal Player*

portalplayer

***

*Embedded Player*

embedplayer


.circleci/config.yml
```diff
@@ -19,7 +19,6 @@ jobs:
       # when movign back to yarn, also the v1-dependencies key should be set back to `yarn-lock`
       # - run: yarn install --pure-lockfile
       - run: npm install
-      - run: npm install web3@1.0.0-beta.29 --save
```
bent0b0x: By adding this dependency explicitly to `package.json/package-lock.json` we don't need the extra step after `npm install`

package.json
```diff
@@ -33,6 +33,11 @@
   },
   "babel": {
     "env": {
+      "development": {
+        "plugins": [
+          "babel-plugin-styled-components"
```
bent0b0x: This puts readable class names in the DOM for development so that we can trace elements back to their components. Example:

image

src/scripts/components/App.js
```diff
@@ -14,27 +14,25 @@ import VideoManager from 'containers/VideoManagerContainer'
...
 
+import { paratiiTheme } from 'constants/ApplicationConstants'
+
 type Props = {
-  match: RouteMatch
+  match: RouteMatch,
+  setSelectedVideo: (id: string) => void
 }
 
-const paratiiTheme = {
```
bent0b0x: This was just a refactor of where the theme lives

src/scripts/components/VideoOverlay.js
```diff
@@ -2,57 +2,148 @@
...
+          const ProfileButton: Class<
+            React.Component<any>
+          > = ((ProfileButtonModule.default: any): Class<React.Component<any>>)
+          this.setState(prevState => ({
+            buttons: {
+              ...prevState.buttons,
+              profile: ProfileButton
+            }
+          }))
+        }
```
bent0b0x: Above is probably the most interesting part of this PR. Basically, this instructs webpack to split `ProfileButton.js` into another module and only load it dynamically when we go down this code path (i.e. when we are viewing the player via the `/embed` route. This prevents us from loading these "plugins" unless they are going to be used.

src/scripts/components/foundations/buttons/IconButton.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
```
bent0b0x: @eliawk @jellegerbrandy @jrgarou I think we should use this for all of our icon buttons through the app. In fact I/we should probably also create a generic `Icon.js` component that this button component uses and would be useful itself in other contexts.

src/scripts/components/widgets/PlayerPlugins/ProfileButton.js
```diff
@@ -0,0 +1,32 @@
...
+  onClick: (e: Object) => void
+}
+
+const ProfileButton = (props: Props) => {
+  const { popoverPortal, popoverOpen, onClick } = props
+
+  return (
+    <Fragment>
+      <IconButton
+        data-test-id="overlay-profile-button"
```
bent0b0x: Given that `styled-components` class names are not something we can rely on in our tests, I propose we start using a `data-test-id` in our components for testing purposes when necessary. Thoughts?

PR #118, opened 2018-02-07
22 user data refactor

bent0b0x:
In starting work on the profile player plugin/popover, I believe I saw a need to refactor how user/login state was managed. Specifically, we need to put more in the `redux` store so that other parts of the app can reliably access things like `walletKey`, `mnemonicKey` (and subsequently get the wallet info from `localStorage`. Perhaps we can eventually move the entire wallet into `redux` and persist it to `localStorage` as well, but I did not take that step in this PR.

@jellegerbrandy @eliawk two things:
1. there is a lot in here, at least at first glance. Let me know if you disagree w/ any decisions made here
2. Can you help debug why the upload tests are failing 🤷‍♂️ ?  The failures do not happen on `dev` so it is definitely the result of one of my code changes, but I did not touch the uploader. The error is:
```shell
Error: No video with id 'CFOc4svIYG6t' was registered
    at ParatiiEthVids.get$ (/home/circleci/repo/node_modules/paratii-lib/dist/paratii.eth.vids.js:155:21)
    at tryCatch (/home/circleci/repo/node_modules/regenerator-runtime/runtime.js:62:40)
```
The UI does report render this id, so perhaps this has uncovered some sort of bug. 

NOTE: please feel free to tell me that everything in this PR is wrong 😀. Perhaps in reality we will always create the wallet on app load or login so we do not need to store any specific wallet information in the store itself, but can always rely on `Paratii-lib` having access to the wallet


src/scripts/actions/UserActions.js
```diff
@@ -4,27 +4,97 @@ import { createAction } from 'redux-actions'
...
+    wallet = paratii.eth.wallet.encrypt(DEFAULT_PASSWORD)
+    dispatch(
+      setAndSyncWalletData({
+        wallet,
+        walletKey,
+        mnemonic,
+        mnemonicKey
+      })
+    )
+  }
```
bent0b0x: @jellegerbrandy I refactored this logic a bit (and moved it to this `UserActions.js` module). 

The idea behind the change was:
- move it here because this is asynchronous logic that affects our state; thus, it should be an action
- this needed to be more connected to our state; now, the action tries to grab the correct keys from the store and if it does not get them, it creates a new wallet and updates the store

src/scripts/components/App.js
```diff
@@ -11,8 +11,7 @@ import PlayContainer from 'containers/PlayContainer'
 import DebugContainer from 'containers/DebugContainer'
 import VideoManager from 'containers/VideoManagerContainer'
 
-import type { RouteMatch } from 'types/ApplicationTypes'
-import { setupKeystore } from 'utils/ParatiiLib'
+import type { Match } from 'react-router-dom'
```
bent0b0x: I realized that `react-router-dom` already had a better typedef for this property than we had hacked in

src/scripts/components/App.js
```diff
@@ -35,8 +35,13 @@ const PortalPlayWrapper = styled.div`
...
 
 class App extends Component<Props, void> {
+  constructor (props: Props) {
+    super(props)
+
+    this.props.initializeApp()
+  }
+
   render () {
-    setupKeystore()
```
bent0b0x: Not good practice to call code w/ side-effects on every render. Better to put this above in the constructor

src/scripts/records/UserRecords.js
```diff
@@ -1,31 +1,31 @@
...
 
-class User extends ImmutableRecord({
-  email: null,
-  name: 'John Doe', // TODO
-  isLoggingIn: false,
-  keepUrl: true
-}) {
-  email: string;
-  name: string;
-  isLoggingIn: boolean;
```
bent0b0x: Keeping things in state like `isLogginIn` are error prone, in my opinion. This should be a calculated piece of state (something more fit for a selector). Thus, I have switched to storing `loginRequestStatus`, updating that via actions, and creating an `isLogginIn` selector that reads this new status property

src/scripts/setupDevEnv.js
```diff
@@ -29,7 +31,19 @@ async function deployContracts () {
 
   const diagnosis = await paratii.diagnose()
   console.log(diagnosis)
-  process.exit(0)
+}
+
+async function seedVideos () {
+  await paratii.core.vids.create({
```
bent0b0x: Now this video is created every time we launch the dev server

src/scripts/selectors/UserSelectors.js
```diff
@@ -1,7 +1,21 @@
...
+  (loginRequestStatus: RequestStatus): boolean =>
+    loginRequestStatus === REQUEST_STATUS.PENDING
+)
 
-export const getIsLoggedIn = createSelector(
-  [getUser, getIsLoggingIn], (user, isLoggingIn) => !!(user && user.get('email') && !isLoggingIn)
+export const getIsLoggedIn: (state: RootState) => boolean = createSelector(
+  [getUser, getIsLoggingIn],
+  (user: User, isLoggingIn: boolean): boolean =>
+    !!(user.get('email') && !isLoggingIn)
```
bent0b0x: @jellegerbrandy I am considering changing this logic to be more like `!!user.get('walletKey') && !isLoggingIn`. My understanding is the presence of a wallet is the true indicator that the user is ready to go. Is that correct?

Or perhaps `isLoggedIn` and `hasWallet` (a selector that doesn't exist yet) should be treated separately. Thoughts?

src/scripts/components/LoginForm.js
```diff
@@ -22,19 +22,19 @@ class LoginForm extends Component<Props, void> {
     return (
       <Form onSubmit={onSubmit}>
         <Input
-          id="login-email"
+          data-test-id="login-email"
```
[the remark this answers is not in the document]
bent0b0x: We definitely don't _need_ it, but it is a convention I _think_ could be helpful. Basically by calling it `[data-test-id]` we indicate that this attribute should be used for nothing other than testing. We will be less inclined to do `document.querySelector('#myId')` (instead of doing something in the react way). Not that there is a high chance someone doing that, but it's just a convention I've recently become aware of and kind of like. But feel free to ditch it :)

src/scripts/actions/UserActions.js
```diff
@@ -4,27 +4,97 @@ import { createAction } from 'redux-actions'
...
+  // TODO: needs to be encrypted
+  localStorage.setItem(mnemonicKey, mnemonic)
+}
+
+export const setupKeystore = () => async (
+  dispatch: Dispatch,
+  getState: () => RootState
+) => {
+  paratii.eth.wallet.clear()
+  const walletKey: string = getWalletKey(getState())
```
bent0b0x: Although refactored, the app still always starts w/ a default key of `keystore-anon`. In reality, though, if the user has secured their account we need some way to get their last wallet key.

However, given how it is currently stored, (where `walletKey` is the actual localStorage key), I'm not sure that would be possible.

Thoughts, @jellegerbrandy, on storing all of our paratii data in `localStorage` in this format?

`localStorage.setItem('paratii', JSON.stringify({ walletKey: 'keystore-anon', wallet: { data: value, key: thing }, //etc }))`

Then on app initialization we can `JSON.parse(localStorage.getItem('paratii'))` one time and put the data in the store. Our defaults will still remain `keystore-anon` and `mnemonic-anon`

bent0b0x:
Thanks for fixing the build!

PR #122, opened 2018-02-07
Uploader ui - Card Component

jrgarou:
- Created [components/foundations/forms/SignForm](https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-99534e5e0cf4bb6337d9983acd5294d2) style(basically for alignment)
- Created [components/foundations/forms/Input.js](https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-4d8727799083f5e3b2b2d74d02ff232d). Just the input element.
- Input focus state style
- Buttons and TextField disabled state
- Created [Card Component](https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-5cd3d0bbb9ab5e3a4c28ad893eb7a867). How it works: https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-51cf2e9651079012ffa2c761d8262281R17
image
- Moved `components/structures/MainHeader.js` to `components/structures/MainHeader/MainHeader.js`, and `components/structures/MainFooter.js` and  to `components/structures/MainFooter/MainFooter.js`. 
- Moved `components/structures/widgets/MainNavigation` to `components/structures/header/MainNavigation`. This components will some widgets or foundations.
- Create `components/pages/Login` and  `components/pages/Signup`, and changed the components at [containers/LoginContainer.js](https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-e6e44fb5ea1cd0365311ca6ba9bdd12c) and  [containers/SignupContainer.js](https://github.com/Paratii-Video/paratii-portal/compare/uploader-ui?expand=1#diff-dbb6768436158ee26ced1c98282e43fd)


src/scripts/components/foundations/forms/SignForm.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components'
```
bent0b0x: I don't think something this specific belongs in `foundations/`. I think a `forms/` directory _outside_ of `foundations/` makes more sense. IMO, `foundations/` should be reserved for very basic building blocks not specific to any particular piece of functionality

src/scripts/components/pages/Login.js
```diff
@@ -0,0 +1,35 @@
...
+
+class Login extends Component {
+  render () {
+    return (
+      <div>
+        <Card
+          title="Login"
+          footer={
+            <FooterText>
+              Don’t have an account?{' '}
```
bent0b0x: Is this `{' '}` space needed?

src/scripts/components/pages/Signup.js
```diff
@@ -0,0 +1,35 @@
...
+
+class Signup extends Component {
+  render () {
+    return (
+      <div>
+        <Card
+          title="Sign up"
+          footer={
+            <FooterText>
+              Already have an account?{' '}
```
bent0b0x: Same do we need this `{' '}`? Could it be taken care of w/ css?

src/scripts/components/structures/Main.js
```diff
@@ -8,11 +8,7 @@ type Props = {
 const MainWrapper = styled.main`
   background-color: ${props => props.theme.colors.body.background};
   flex: 1 1 100%;
-  display: flex;
   padding: 80px;
-  overflow-x: hidden;
-  height: 100vh;
-  justify-content: center;
```
bent0b0x: Why were these removed? I did see lots of issues before adding this (like you could scroll strangely in the x direction and the page height was odd. Does the app perform correctly at all viewport dimensions now?

src/scripts/components/structures/header/MainHeader.js
```diff
@@ -1,8 +1,8 @@
 import React, { Component } from 'react'
 import styled from 'styled-components'
-import MainHeaderLogo from '../widgets/MainHeaderLogo'
-import SearchInput from '../widgets/SearchInput'
-import MainNavigation from '../widgets/MainNavigation'
+import MainHeaderLogo from 'components/widgets/MainHeaderLogo'
+import SearchInput from 'components/widgets/SearchInput'
+import MainNavigation from 'components/structures/header/MainNavigation'
```
bent0b0x: 🎉

src/scripts/components/widgets/FilesUploader.js
```diff
@@ -2,7 +2,7 @@ import React, { Component } from 'react'
 import styled, { css } from 'styled-components'
 import FilesUploaderSvg from '../foundations/svgs/FilesUploaderSvg'
 import Button from '../foundations/buttons/Button'
-import TextField from '../foundations/forms/TextField'
+import TextField from '../widgets/forms/TextField'
```
bent0b0x: Not a big deal but this file is already in `widgets` so this can be `./forms` or just `components/widgets/forms/etc`

src/styles/app.scss
```diff
@@ -5,8 +5,17 @@ html, body, #root {
   width: 100%;
 }
 
+#root {
+  overflow-x: hidden;
```
bent0b0x: Ah, I see this is where you moved the changes I referenced a few comments earlier :)

bent0b0x:
@jellegerbrandy @jrgarou after our discussion earlier today on gitter, is there even a use-case for logging in w/ email and password anymore? My understanding (which is perhaps still incorrect) is that the user cannot log in w/ email. They can only decrypt their wallet via a password. Is that correct?

src/scripts/components/foundations/forms/SignForm.js
jrgarou: I totally agree, @bent0b0x.
They are foundations form elements: Input and SignForm. Both just styled-components with no functionality. The `forms/` directory It's just a way to organize those components inside `foundations/`.
On `components/widgets/forms/TextField.js` we got our input with some functionality.

On the same `foundations/` directory we got `buttons/`, `svgs/` and `base/`. 
I think we could just move `base/` directory (with some settings of animation, colors, sizes, typography and themes) to outside the components.

What you think?

src/scripts/components/pages/Login.js
jrgarou: Removing this...

src/scripts/components/pages/Signup.js
jrgarou: Removing this...

src/scripts/components/foundations/forms/SignForm.js
bent0b0x: Yes, I think moving the settings for animations, colors, etc makes sense _outside_ of `components/`. Maybe in `constants`/ as `AnimationConstants.js`, `ColorsConstants.js`, etc.

And as for where `SignForm.js` should go, I think the only thing that made me question it was the name. Might it be possible for us to rename this just `Form` (if it is generic enough to use in other contexts)? The way I think about it is everything in `foundations/` could, theoretically, be pulled out and used as a UI component library to build other apps in the future; so its components should be fairly generic and not specific to any particular application logic.

Again, this part isn't really that important so let's not let it hold us up 😄

PR #128, opened 2018-02-09
22 - store wallet in state

bent0b0x:
This PR puts the wallet in an immutable record in `redux`. This gives us reliable access to things like the user's address so that we can make calls to `paratii-lib` re: their balance and other account information.

Once these changes are in `dev`, I will start on the code that will fetch and store the user's balance, and then render it in the UI.

### DIFF
--- a/src/scripts/actions/UserActions.js
+++ b/src/scripts/actions/UserActions.js
@@ -47,14 +47,16 @@ const setAndSyncWalletData = ({
   mnemonicKey
 }: {
   walletKey: string,
-  wallet: Object,
+  wallet: Array<Object>,
   mnemonicKey: string,
   mnemonic: string
 }) => (dispatch: Dispatch): void => {
   dispatch(
     setWalletData({
       walletKey,
-      mnemonicKey
+      mnemonicKey,
+      wallet: (wallet || [])[0] || {},
+      mnemonic
     })
   )
   localStorage.setItem(walletKey, JSON.stringify(wallet))
@@ -71,13 +73,15 @@ export const setupKeystore = () => async (
   const walletKey: string = getWalletKey(getState())
   const mnemonicKey: string = getMnemonicKey(getState())
   const walletString: string = localStorage.getItem(walletKey) || ''
-  let wallet: ?Object
+  let decryptedWallet: ?Object
   let mnemonic: string = localStorage.getItem(mnemonicKey) || ''
   let walletIsValid: boolean = true
+  let walletToSave: ?Array<Object>
   if (walletString) {
     try {
-      wallet = paratii.eth.wallet.decrypt(
-        JSON.parse(walletString),
+      walletToSave = JSON.parse(walletString)
+      decryptedWallet = paratii.eth.wallet.decrypt(
+        walletToSave,
         DEFAULT_PASSWORD
       )
       paratii.eth.setAccount(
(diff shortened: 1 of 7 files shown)


src/scripts/actions/UserActions.js
```diff
@@ -88,17 +92,17 @@ export const setupKeystore = () => async (
...
-      setAndSyncWalletData({
-        wallet,
-        walletKey,
-        mnemonic,
-        mnemonicKey
-      })
-    )
+    walletToSave = paratii.eth.wallet.encrypt(DEFAULT_PASSWORD)
   }
+  dispatch(
```
bent0b0x: The changes here boil down to this:

Whether we find a valid wallet in local storage or create a new one, we dispatch an action that will save that wallet in the `redux` store

src/scripts/records/UserRecords.js
```diff
@@ -3,29 +3,106 @@
...
+}) {
+  version: number
+  id: string
+  address: string
+  crypto: Crypto
+
+  constructor ({ crypto, ...rest }: Object = {}) {
+    super({ ...rest, crypto: new Crypto(crypto) })
+  }
+}
```
bent0b0x: This record (and the ones above it) match the shape of the wallet created by `paratii-lib` and stored in `localStorage`

src/scripts/records/UserRecords.js
```diff
@@ -3,29 +3,106 @@
...
+  address: '',
+  crypto: new Crypto()
+}) {
+  version: number
+  id: string
+  address: string
+  crypto: Crypto
+
+  constructor ({ crypto, ...rest }: Object = {}) {
+    super({ ...rest, crypto: new Crypto(crypto) })
```
bent0b0x: This technique is useful because it allows you to pass a a plain JS object to the constructor of an immutable record and that record will "automatically" convert pieces of that object to the appropriate records.

If we do not do this, which is also a valid approach, then we would have to pass in a JS object that already has certain keys set to the appropriate immutable records.

src/scripts/selectors/UserSelectors.js
```diff
@@ -19,3 +19,6 @@ export const getIsLoggedIn: (state: RootState) => boolean = createSelector(
   (user: User, isLoggingIn: boolean): boolean =>
     !!(user.get('email') && !isLoggingIn)
 )
+
+export const getAddress = (state: RootState): string =>
+  createSelector([getWallet], (wallet: Wallet): string => wallet.get('address'))
```
bent0b0x: Here is the real payoff for this PR 😄 .

Now we can use this selector to get the wallet address and make a call to `paratii-lib` to fetch, for example, PTI balance

jellegerbrandy:
About the security considerations: these data that are in saved in the redux state with this PR are also saved in localStorage, so this PR does not make it less secure.

That said, this is somethign that does need to be looked at. 
Referring to the specs here: https://github.com/Paratii-Video/paratii-portal/wiki/Portal-Specs:-wallet-handling#securing-and-personalizing-the-account

So there is a sort fo natural distincition between two ways of interacting with the app:

- informal "anonymous" usage - the application has full control over the wallet, and has access to all the secrets (password with which the wallet is encrypted, mneomnic seed phrase). **this is what we have now**
- secured, personalized usage: the user has been prompted and has taken action to secure his account: he has chose a password for encrypting the wallet, and has declared that he has saved the mnemonic). After this, we are (of course) not saving these secrets anywhere. This we do not have yet.

src/scripts/selectors/UserSelectors.js
jellegerbrandy: Ok, butbut  have `paratii.config.accounts.address` already, which must be set for paratii-lib to function, and which will **whatever is in the redux state**, be the address with wich transactions are signed, etc.

bent0b0x: image

Then why do we need to pass in an address to this call?

jellegerbrandy: because we may be interested in the balances of other accounts as well as our own. (In theory, I have no use case yet, but you could imagine stuff like "I'm only sending you my video if I know you can pay for it).

Issue #115, opened 2018-02-06
Create voucher component (and page)


jellegerbrandy:
This is a simple form that has a single input field for inputting a string, and a sumbit button ("redeem voucher"). 

It should call `paratii.eth.voucher.redeem(myvoucherstring)`, which will returna  promise. Success also means that the balance of the user is updated.

For now, we can show this on a route to `/voucher`.

See https://github.com/Paratii-Video/paratii-lib/issues/103

jrgarou:
@eliawk the component: https://github.com/Paratii-Video/paratii-portal/commit/3fb6cb6cb34a3fc1007c0d35ab14ac12fe4764b5

jellegerbrandy:
It's working!

PR #150, opened 2018-02-12
added uploader and transcoder update

eliawk:


src/scripts/components/VideoForm.js
```diff
@@ -118,6 +123,30 @@ class VideoForm extends Component<Props, Object> {
       title: nextProps.selectedVideo.title,
       description: nextProps.selectedVideo.description
     })
+
```
eliawk: I save in the component state the value coming form the global state. Don't know if it's a bad approach..

src/scripts/components/VideoListItem.js
```diff
@@ -72,19 +72,53 @@ const Bar = styled.div`
...
+      totalProgress: 0
+    }
     this.handleClick = this.handleClick.bind(this)
   }
 
   handleClick () {
     this.props.onClick(this.props.video.id)
   }
 
+  componentWillReceiveProps (nextProps: Props): void {
```
eliawk: this code is the same of `VideoForm.js`, this could be written once? by the way there is an issue, because when the file has been transcoded is correctly updated in `VideoForm.js` but not in `VideoListItem.js`

jellegerbrandy: ok can you create an issue for the issue? :-)

jellegerbrandy: (or fix it of course :-))

jrgarou:
the voucher component is in this PR(https://github.com/Paratii-Video/paratii-portal/issues/115)
https://github.com/Paratii-Video/paratii-portal/pull/150/commits/3fb6cb6cb34a3fc1007c0d35ab14ac12fe4764b5

Nothing special. Just a `<Card />` with some others components

Issue #160, opened 2018-02-12
Review reducers and ensure they always return fully immutable state


bent0b0x:
We have some examples of reducers that are creating non-immutable pieces of substate. We need to keep the full-state immutable.

Example:

https://github.com/Paratii-Video/paratii-portal/blob/134-notfound/src/scripts/reducers/VideosReducer.js#L159-L170

This should instead be something like the following 👇 

```javascript
state = state.mergeDeep({
      [payload.id]: {
        ipfsHash: ipfsHash,
        transcodingStatus: new AsyncTaskStatusRecord({
          name: 'success',
          data: new SomeRecord({
            ipfsHash: ipfsHash,
            sizes: payload.sizes
          })
        })
      }
    })
```

cc @eliawk @jellegerbrandy

bent0b0x:
@eliawk I'm going to try and tackle this one now

bent0b0x:
I think my initial assessment of the problem here was wrong. The state does always stay `immutable`. However, there are some issues with how `.mergeDeep` is used (I think), in that sometimes data is merged into nested immutable objects when in fact they should be completely overwritten.

bent0b0x:
@eliawk I'm pretty sure I was wrong about this issue to begin with. I think there are some parts of the video reducers that could be _slightly_ improved, but not what I reported here. My bad!

I will tackle those other issues via #112

PR #162, opened 2018-02-13
160 - restructure some reducers and add some tests

bent0b0x:
**Issue**
#160 

**Work Done**
- rewrite some reducers to ensure they do not merge nested data when they should in fact completely overwrite the nested data
- add some reducer unit tests


bent0b0x:
Closing this for now. I will take a TDD approach and write the unit tests _before_ modifying the reducers. This will clear some things up!

PR #165, opened 2018-02-13
8 uploader

eliawk:


test/functional-tests/uploader.spec.js
```diff
@@ -32,6 +32,7 @@ describe('Uploader Tool', function () {
     assert.isOk(videoId)
     assert.isOk(videoId.length > 8)
     // set title and video in the form
+    browser.pause(1000)
```
bent0b0x: Let’s work on an alternative solution. Pause killed us in the last app 😅

eliawk: yes i know, was just a temp solution. i agree with you!

src/scripts/actions/UploadActions.js
```diff
@@ -47,7 +53,9 @@ export const upload = (file: Object) => (dispatch: Dispatch<*>) => {
     dispatch(uploadProgress({ id: newVideoId, progress: progressPercent }))
   })
   uploader.on('fileReady', function (file) {
-    dispatch(uploadLocalSuccess({ id: newVideoId, hash: file.hash }))
```
bent0b0x: Might want to set a default value for file just in _case_ it comes back `null/undefined`. Or just check that it is truthy before grabbing properties off of it to avoid a runtime error

src/scripts/components/VideoForm.js
```diff
@@ -172,6 +176,8 @@ class VideoForm extends Component<Props, Object> {
 
   handleSubmit (e: Object) {
     e.preventDefault()
+    console.log('!!!!!  handlesubmit')
+    console.log(this.state.title)
```
bent0b0x: Remove logs please :)

src/scripts/components/VideoForm.js
```diff
@@ -155,13 +159,13 @@ class VideoForm extends Component<Props, Object> {
...
-      this.setState({
-        totalProgress: Math.round(
-          (this.state.uploadProgress + this.state.transcodingProgress) / 2
-        )
-      })
     }
+
+    this.setState((prevState, nextProps) => ({
+      totalProgress: Math.round(
+        (prevState.uploadProgress + prevState.transcodingProgress) / 2
```
bent0b0x: Is `uploadProgress` something that is originally coming from props/redux store? If so, I think we'll want to change this to be a selector. It could go in an `UploadSelectors.js` module (new if it doesn't exist). 

Selectors are the place where any sort of complex (or not very very simple) logic that process/transforms/combines store data should live, IMO. Also, if this data is coming from props then there is no need to duplicate it in state.

src/scripts/components/VideoForm.js
```diff
@@ -182,11 +188,11 @@ class VideoForm extends Component<Props, Object> {
 
   render () {
     const video: ?VideoRecord = this.props.selectedVideo
-    // const uploadProgress = video.getIn(['uploadStatus', 'data', 'progress'])
-    // const transcodingProgress = video.getIn(['transcodingStatus', 'data', 'progress'])
     const thumbImages =
       video &&
       video.getIn(['transcodingStatus', 'data', 'sizes', 'screenshots'])
+    const fileSize = prettyBytes((video && video.get('filesize')) || 0)
+    console.log(video)
```
bent0b0x: Remove logging

src/scripts/components/VideoForm.js
```diff
@@ -196,6 +202,9 @@ class VideoForm extends Component<Props, Object> {
     } else {
       thumbImage = 'http://paratii.video/public/images/paratii-src.png'
     }
+    // TODO get the correct image form transcoder, now hardocoded
+    // thumbImage = 'http://paratii.video/public/images/paratii-src.png'
```
bent0b0x: Remove comments?

src/scripts/components/VideoForm.js
```diff
@@ -204,15 +213,15 @@ class VideoForm extends Component<Props, Object> {
           <Hidden>
             ({this.state.id} - {ipfsHash})
           </Hidden>{' '}
-          <VideoFormSubTitle purple>345MB</VideoFormSubTitle>
+          <VideoFormSubTitle purple>{fileSize}</VideoFormSubTitle>
```
bent0b0x: Yay for removing hard-coded values!

src/scripts/components/VideoListItem.js
```diff
@@ -103,21 +103,15 @@ class UploadListItem extends Component<Props, void> {
       this.setState({ transcodingProgress: 100 })
     }
 
-    this.setState({
+    this.setState((prevState, nextProps) => ({
       totalProgress: Math.round(
-        (this.state.uploadProgress + this.state.transcodingProgress) / 2
+        (prevState.uploadProgress + prevState.transcodingProgress) / 2
```
bent0b0x: Looks like this is duplicated in two components. If extracted to a selector, then we have no problem. The transformed data can be passed to both components w/ ease.

src/scripts/components/VideoListItem.js
```diff
@@ -103,21 +103,15 @@ class UploadListItem extends Component<Props, void> {
...
       )
-    })
+    }))
   }
 
   render () {
     const item = this.props.video
-    // let progress
```
bent0b0x: Remove these?

src/scripts/reducers/VideosReducer.js
```diff
@@ -27,7 +27,7 @@ import type { Action, VideoRecordMap } from 'types/ApplicationTypes'
 const reducer = {
   [UPLOAD_REQUESTED]: (
     state: VideoRecordMap,
-    { payload }: Action<{ id: string, filename: string }> = {}
```
bent0b0x: I would keep the `= {}` default just in case we call the action w/ no payload (to avoid runtime error)

src/scripts/selectors/index.js
```diff
@@ -18,6 +18,7 @@ import type { RootState, RequestStatus } from 'types/ApplicationTypes'
 export const getVideo = (state: RootState): ?VideoRecord => {
   const videoId: ?string = state.selectedVideo
   const video = videoId && state.videos.get(videoId)
+  console.log('selector', video)
```
bent0b0x: Remove logging

src/scripts/utils/AppUtils.js
```diff
@@ -46,3 +46,15 @@ export const getParatiiConfig = (env: ?string = 'development'): Object => {
 
   return config
 }
+
+export const prettyBytes = (num: ?number): string => {
+  if (num === null || num === undefined) {
+    num = 0
+  }
+  const UNITS = ['B', 'kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
```
bent0b0x: Can these be moved to `ApplicationConstants.js` and imported into `AppUtils.js`? It is possible we may want to use this list elsewhere. Not a big deal though, can always move later.

test/functional-tests/player.test.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs')
 const Promise = require('bluebird')
 const path = require('path')
 
-describe('Player:', function () {
+describe('🎥 Player:', function () {
```
bent0b0x: 😎

test/functional-tests/player.test.js
```diff
@@ -87,6 +87,7 @@ describe('Player:', function () {
   describe('portal player', () => {
     it('plays a video automatically', () => {
       browser.url(`http://localhost:8080/play/${videoId}`)
+      browser.pause(1000)
```
bent0b0x: Can this be rewritten w/o pause? I don't think I've seen issues with this test failing

test/functional-tests/player.test.js
```diff
@@ -113,6 +114,7 @@ describe('Player:', function () {
   describe('embedded player', () => {
     it('plays a video automatically', () => {
       browser.url(`http://localhost:8080/embed/${videoId}`)
+      browser.pause(1000)
```
bent0b0x: Remove this?

test/functional-tests/setup.spec.js
```diff
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { paratii } from './test-utils/helpers'
+// import { paratii } from './test-utils/helpers'
```
bent0b0x: Remove comments?

Issue #166, opened 2018-02-13
avoid test timeout in circleci


jellegerbrandy:
Sometimes, the fisrt test fails because the server is simply not responding yet.

Proposed solution:

add a step in circleci/config.yml before `yarn run chimp-test` that waits until the server responds on port 8080.

jellegerbrandy:
Fixed by #167

PR #167, opened 2018-02-14
112 - video reducer unit tests (and a few other things)

### LINKED ISSUE #112: Add some reducer unit tests
Let's start with `VideoReducer.js`

bent0b0x:
**Work Done**
- add unit tests for all actions handled by the videos reducer
- remove a reducer path that wasn't being used
- change/fix some logic in the reducers (with a focus on ensuring edge cases are covered and that the whole state remains immutable)
- add a scripts/configure circle so that chimp tests do not start until dev server is responsive

@jellegerbrandy I changed code in all of the video reducer code paths, so feel free to do some testing to make sure nothing is screwed up. I do not know the uploader as well at this point to be able to tell for sure that things are working 100% as designed. (Also I think the transcoder is still not working so difficult to test the entire workflow).

Resolves #112, #160, and #166

### DIFF
--- a/src/scripts/components/VideoForm.js
+++ b/src/scripts/components/VideoForm.js
@@ -16,7 +16,7 @@ import VideoProgress from 'components/widgets/VideoForm/VideoProgress'
 import Hidden from 'components/foundations/Hidden'
 
 type Props = {
-  selectedVideo: VideoRecord,
+  selectedVideo: ?VideoRecord,
   canSubmit: boolean,
   saveVideoInfo: Object => Object
 }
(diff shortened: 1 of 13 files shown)


src/scripts/components/VideoForm.js
```diff
@@ -117,36 +117,51 @@ class VideoForm extends Component<Props, Object> {
   }
 
   componentWillReceiveProps (nextProps: Props): void {
-    this.setState(nextProps.selectedVideo)
-    this.setState({
-      id: nextProps.selectedVideo.id,
-      title: nextProps.selectedVideo.title,
-      description: nextProps.selectedVideo.description
-    })
+    const { selectedVideo } = nextProps
```
bent0b0x: The changes in non-reducer files were mainly bugs that were uncovered during work on this PR. Mostly that some props could be `undefined/null` in components and that case was not accounted for in the components themselves.

src/scripts/waitForServer.js
```diff
@@ -0,0 +1,27 @@
...
+        process.exit(1)
+      }
+    }
+  )
+
+  req.on('error', pingServer)
+  req.end()
+}
+
+pingServer()
```
bent0b0x: Here's the hacky little util I wrote to not start our chimp tests until the server responds. So far, it seems to be working!

test/unit-tests/reducers/VideosReducer.test.js
```diff
@@ -0,0 +1,1588 @@
+/* @flow */
```
bent0b0x: Enjoy reading through this diff...

PR #171, opened 2018-02-16
163 modal

jrgarou:
### How to use
The modal is set up on [App.js](https://github.com/Paratii-Video/paratii-portal/compare/163-modal?expand=1#diff-78b6cff1445139b5ae07e917ca2dd7e3). 
To call the modal it need to pass the `App.js`(`showModal={this.showModal}`) [method](https://github.com/Paratii-Video/paratii-portal/compare/163-modal?expand=1#diff-78b6cff1445139b5ae07e917ca2dd7e3R56) to the _page_

This is how it works on `Home` page:

- on [App.js](https://github.com/Paratii-Video/paratii-portal/compare/163-modal?expand=1#diff-78b6cff1445139b5ae07e917ca2dd7e3)
```
import Modal from './widgets/modals/Modal'
...
constructor (props: Props) {
...
    this.state = {
      modalContent: Object,
      showModal: false
    }

    this.showModal = this.showModal.bind(this)
    this.closeModal = this.closeModal.bind(this)
  }

  showModal (View) {
    this.setState({
      modalContent: View,
      showModal: true
    })
  }

  closeModal (view) {
    this.setState({
      showModal: false
    })
  }
...
render () {
  return (
    <Modal
      show={this.state.showModal}
      closeModal={this.closeModal}
      content={this.state.modalContent}
    />
...
    <Route
      exact
      path="/"
      render={props => <Home showModal={this.showModal} />}
    />
```

- on [home](https://github.com/Paratii-Video/paratii-portal/compare/163-modal?expand=1#diff-d14de28fca6c7124f2d50bfad1cdbc07R23)
```
import ModalStake from 'components/widgets/modals/ModalStake

  <Btn
    onClick={() => {
      this.props.showModal(<ModalStake />)
    }}
    purple
  >
```

@bent0b0x What you think about this approach?


src/scripts/components/App.js
```diff
@@ -42,17 +43,48 @@ class App extends Component<Props, void> {
...
+          />
+
           <MainHeader />
           <Main>
             <Switch>
-              <Route exact path="/" component={Home} />
+              <Route
+                exact
+                path="/"
+                render={props => <Home showModal={this.showModal} />}
```
bent0b0x: If I understand correctly, does this mean when we want a component to show a modal we need to explicitly pass a `showModal` prop to it?

I think if we put what modal content we are showing in `redux`, we could avoid this. We could have a piece of state called `modal`, and have actions that would update that piece of state (either to set modal content or clear it). Then, any `connect`-ed component could dispatch one of those actions w/o us having to explicitly pass in a `showModal` prop. The `<Modal />` component would then also be `connect`-ed to `redux` and whenever the modal state indicated that it had modal content to show, it would render it. 

Thoughts on this?

src/scripts/components/foundations/Title.js
```diff
@@ -1,16 +1,5 @@
...
-  display: flex;
-  align-items: center;
-  color: #818181;
-  font-weight: 600;
-  margin-bottom: 15px;
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  width: 100%;
+  color: white;
```
bent0b0x: Can we use one of our defined color variables here?

src/scripts/components/foundations/svgs/MainSvg.js
```diff
@@ -72,6 +72,9 @@ class MainSvg extends Component<Props, void> {
             fill="#fff"
           />
         </symbol>
+        <symbol id="icon-close" viewBox="0 0 512 512">
```
bent0b0x: Can we add new SVGs to `/assets/img`? I'm not sure I like the idea of building a giant SVG that will be loaded into the DOM on every render.

jrgarou: This is only for reusable svgs. It doesn't need to make a request for every icon. It's pretty small, in fact

src/scripts/components/foundations/Title.js
jrgarou: It can. This is not defined yet

src/scripts/components/foundations/svgs/MainSvg.js
bent0b0x: Gotcha. Maybe in the future it could grow very large but fine for now. 

My only gripe now is that we are loading/managing SVGs in two different ways now, so we’ll want to reconcile that.

jrgarou: The MainSvg holds the icons basically. The others svgs are set directly in the html.

src/scripts/components/App.js
bent0b0x: Were these changes made/will they be?

PR #187, opened 2018-02-20
185 - indicate which video is selected in uploader

bent0b0x:
**Issue**
#187 

Indicate in the Uploader UI which video is selected.

selected


pedrocasa:
Hi Ben, 
this light grey rectangle was designed to indicate which video you are selecting

capture

bent0b0x:
Got it, @pedrocasa. Right now, we are using that light gray for every other video in the list to make the list look prettier. Are you okay with just scrapping that and giving all videos here (except selected one) same background?

eliawk:
@ben this was the original @pedrocasa idea, using the light gray to display the selected video, but this was misunderstood and has became the `:odd` list item color

bent0b0x:
@eliawk @pedrocasa changed the color (see screenshot above)

Issue #188, opened 2018-02-20
Completely separate player state from uploader state


bent0b0x:
`selectedVideo` should not be related to both the player and the uploader. This will (and I think already is) causing bugs.

bent0b0x:
This will require us to remove `selectedVideo` from the top level of state. Instead, it should live _both_ within `state.player` and within a new `state.uploader`. `state.videos` can remain the same, as this will be used in the uploader and elsewhere.

PR #192, opened 2018-02-20
188 - separate the uploader state from the player state

bent0b0x:
**Issue**
#188 

**Problem**
The uploader and the player were sharing state in the form of `state.selectedVideo`. This is error prone and could lead to bugs (and I believe it was already). Instead, the uploader and the player should have entirely separate concepts of selected video.

**Work Done**
- delete `state.selectedVideo`
- add `selectedVideoId` to both the (new) `state.uploader` and `state.player`


eliawk:
this looks like to me, i think we should merge it.

PR #202, opened 2018-02-21
enabling uglify mangle wtih reserved words

ya7ya:
this reduces the bundle.js size from 8.5MB to 6.0MB , I'm still working on getting `compress` to work 😠


webpack.config.js
```diff
@@ -123,7 +123,17 @@ const config = {
         sourceMap: false, // this is an effor to save some memory
         uglifyOptions: {
           ecma: 6,
-          mangle: false,
+          // mangle: false,
```
bent0b0x: Mind removing this comment?

PR #222, opened 2018-02-22
Fix/video duration

eliawk:


src/scripts/actions/UploaderActions.js
```diff
@@ -137,7 +137,7 @@ export const saveVideoInfo = (videoInfo: Object) => async (
     // dispatch(selectVideo(videoInfo.id))
   }
   dispatch(videoDataStart(videoInfo))
-
+  console.log(videoInfo)
```
bent0b0x: Mind removing this logging statement?

PR #234, opened 2018-02-23
Revert "update paratii-lib"

bent0b0x:
This reverts commit c6a06adffb8af27993d2409ac5a4207546185355. The new `paratii-lib` version was breaking uploads w/ this error: 

```
paratii.core.vids.js?6d18:85 Uncaught (in promise) Error: No video to update
    at ParatiiCoreVids.update (paratii.core.vids.js?6d18:85)
    at <anonymous>
```

cc @geckoslair

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8829,17 +8829,6 @@
         "rimraf": "2.6.2"
       }
     },
-    "fstream-ignore": {
-      "version": "1.0.5",
-      "resolved": "https://registry.npmjs.org/fstream-ignore/-/fstream-ignore-1.0.5.tgz",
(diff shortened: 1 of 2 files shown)


jellegerbrandy:
but that error was a new but _good_ error, not a problem with paratii-lib, I think. It is a call to "update" heppening when there was no video to update. (you should call `create` or `upsert`in this case)

bent0b0x:
Possibly, but had to revert for quick fix. I did try something along the lines of what you were saying but still got errors; will have to try again. I recall seeing `get`s fail despite the correct Id being passed in

PR #235, opened 2018-02-23
Load videos

bent0b0x:


bent0b0x:
@eliawk if you have a sec can you merge `dev` into this branch? I think that would pass the build

bent0b0x:
Actually not entirely true, unit tests need to be updated as well. But I think this needs `dev` merged into it anyway

eliawk:
@bent0b0x i see the layout kind broken
schermata 2018-02-23 alle 12 57 30

the video list at the first upload it's gone away

bent0b0x:
@eliawk are you sure this issue is not happening on `dev` too? I do not see how this would be related or caused by this PR.

src/scripts/records/VideoRecords.js
```diff
@@ -16,6 +16,9 @@ class Video extends ImmutableRecord({
   price: '',
   thumbnailUrl: '',
   title: '',
+  author: '',
+  free: '',
```
bent0b0x: I feel like we shouldn't have a `free` field. Doesn't that just mean `price === '0'`? This seems like a duplication of state and we should avoid

Issue #216, opened 2018-02-22
add "author" field to upload form


jellegerbrandy:
This will need a change in paratii-lib as well: https://github.com/Paratii-Video/paratii-lib/issues/136

felipegaucho:
@jellegerbrandy feel free to open it again if the work on the lib is not fully done for this.

Issue #184, opened 2018-02-19
Design.Portal Uploader corrections


pedrocasa:
Here are some corrections for the https://staging.paratii.video/upload

The objective of this issue is to standardize our current page with the UI kits/ Web flows previously designed and

- First of all, we should use the same texts of the last comment of the issue https://github.com/Paratii-Video/paratii-portal/issues/60

Landing page

- The Cards corners need to be rounded. In the software that I'm using it's 4px rounded, but if this unit won't match visually with the front end code we may need to change this value
- The illustration of upload box needs to be updated for this one  ( link for all illustrations indicated by name of the card https://we.tl/r3kuENaQxz )
- The Input field of the forms need to be aligned to the left
- This button is not needed now and may not be needed later. If the link is correct the uploader should automatically go to the "edit" page.
- The last line of footer text of the voucher is the only text that is aligned to the right ( In the current UI kit the only texts that are aligned to the right are the support texts of the forms)
- The subtitle of the PTI guide is missing
- Need to apply the "disactivated button color" to the left arrow button of the PTI guide
- The Roboto bold font should only be used in the buttons, All other texts should be Roboto Regular
- In the last designs that I made I added a new background to the cards (just a small detail). Here is the link https://we.tl/ez1dFHRziV
- (This may be corrected when we disappear with the upload button but I'm gonna mention it here anyway) The footer of the voucher and the upload cards should have the same height.
- The footer text "Discover all about Paratii on paratii.video" should be aligned with the Paratii logo (header).

Edit Page.
- The Cards corners need to be rounded. In the software that I'm using it's 4px rounded, but if this unit won't match visually with the front end code we may need to change this value
- The Roboto bold font should only be used in the buttons, All other texts should be Roboto Regular.
- The footer text "Discover all about Paratii on paratii.video" should be aligned with the Paratii logo (header).
- The Save Data Button should be aligned with add the add another video button
- The video list footer is too big, it should be shorter
- If the Free option is the only one available right now it should already appear selected
- The transcoding loading bar is thicker than the others.
- The loading percentage of loading bars are not aligned to the other elements when it's 0%.
- The " Uploading" "Transcoding" "Data" are too big, it should have the same font size of the loading percentage
- all the info on the right side ( upload/transcoding info, fonts, bars ) is too big, it's consuming all the space.
- The name of the video still the same on the video list and header after I change it.
- Don't need to use "success" after the video uploaded we just need to show the "check" icon and the word Uploaded

[...]

PR #236, opened 2018-02-23
Uploader ui

jrgarou:
- fix responsive uploader issues 
- change FilesUploader content align
- add author field to upload form #216
- Design.Portal Uploader corrections #184


src/scripts/components/VideoList.js
```diff
@@ -49,7 +50,7 @@ class VideoList extends Component<Props, void> {
       ''
     )
     return (
-      <Card margin="0 25px 0 0" nopadding footer={footer}>
+      <Card {...this.props} nopadding footer={footer}>
```
bent0b0x: What props are being passed here? I think it is preferable to not just pass through all props because it is not entirely clear to others what is being used/what is just extra. Can you pass explicit props?

src/scripts/components/widgets/RedeemVoucher.js
```diff
@@ -67,13 +65,18 @@ class RedeemVoucher extends Component<Props, Object> {
   render () {
     return (
       <Card
-        margin={this.props.margin}
+        {...this.props}
```
bent0b0x: Same question w/ passing in all props

src/scripts/components/VideoList.js
jrgarou: > What props are being passed here?

All (and only) `Cards` props. 

Do we need to pass even if we don't use not of them?
Example:
No props are used on Home
https://github.com/Paratii-Video/paratii-portal/blob/dev/src/scripts/components/pages/Home.js#L9

Using marginLeft at VideoManagerContainer
https://github.com/Paratii-Video/paratii-portal/blob/uploader-ui/src/scripts/containers/VideoManagerContainer.js#L54

bent0b0x: I'm not exactly sure what you mean. All I am saying here is that instead of `{...this.props}` we should do `<Component prop1={value} prop2={value2} />`, otherwise other devs cannot see what the props are w/o inspecting the code at runtime.

And if there are props not being used for sure you do not need to pass them. That is always true for any component.

PR #251, opened 2018-02-24
split mediaplayer code for lazy loading

ya7ya:
this introduces the concept of code-splitting explain https://reactjs.org/docs/code-splitting.html
This is useful to keep the initial needed code small enough.
as an exaxmple, I chose to separate the `paratii-mediaplayer` till it's needed to play a video. check out the article above and the code to get an idea of the change required.

__note__ this is also possible for webpacked libs , not just react ( `paratii-lib` ). checkout https://webpack.js.org/guides/code-splitting/ for more info.

### DIFF
--- a/src/scripts/components/Play.js
+++ b/src/scripts/components/Play.js
@@ -14,7 +14,6 @@ import NotFound from './pages/NotFound'
 
 import type { ClapprPlayer } from 'types/ApplicationTypes'
 import type { Match } from 'react-router-dom'
-const CreatePlayer = null // eslint-disable-line no-unused-vars
 
 type Props = {
   match: Match,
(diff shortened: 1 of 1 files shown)


ya7ya:
Okay, i'm not gonna debug flow 😏 ..

jellegerbrandy:
Ah, this is actually very straightfoward.
i think also @bent0b0x has used a pattern similar like this.

src/scripts/components/Play.js
```diff
@@ -3,7 +3,7 @@
 import React, { Component } from 'react'
 import { Events } from 'clappr'
 import styled from 'styled-components'
-import CreatePlayer from 'paratii-mediaplayer'
+// import CreatePlayer from 'paratii-mediaplayer'
```
bent0b0x: Remove comment

jellegerbrandy:
sorry about three flow errors - you forked from a broken branch. Should be fixed now

PR #260, opened 2018-02-26
embedding headers in play and embed route

geckoslair:


parity-config.toml
```diff
@@ -7,5 +7,5 @@ max_peers = 0
 [mining]
 reseal_min_period = 0
 [rpc]
-cors = "*"
+cors = ["*"]
```
bent0b0x: Is this going to work with the version of parity everyone has installed?

src/scripts/components/Play.js
```diff
@@ -319,37 +319,37 @@ class Play extends Component<Props, State> {
   facebook () {
     if (this.props.video) {
       var baseurl = 'https://www.facebook.com/sharer/sharer.php?u='
-      return baseurl + this.portalUrl() + '/embed/' + this.props.video.id
```
bent0b0x: Can you move `portalUrl` to maybe `AppUtils.js`?

src/scripts/components/Play.js
```diff
@@ -319,37 +319,37 @@ class Play extends Component<Props, State> {
...
+      return baseurl + this.portalUrl() + '/play/' + this.props.video.id
     }
   }
   twitter () {
     if (this.props.video) {
       var baseurl = 'https://twitter.com/intent/tweet'
```
bent0b0x: Can you prefer `let/const`? (i.e. never `var` 😄 )

src/scripts/components/Play.js
```diff
@@ -319,37 +319,37 @@ class Play extends Component<Props, State> {
...
-      var url = '?url=' + this.portalUrl() + '/embed/' + this.props.video.id
+      var url = '?url=' + this.portalUrl() + '/play/' + this.props.video.id
       var text = '&text=🎬 Worth a watch: ' + this.props.video.title
       return baseurl + url + text
     }
   }
   whatsapp () {
```
bent0b0x: Can you rename all of these methods to be more clear? I think a better name could be `getWhatsAppUrl`

PR #262, opened 2018-02-26
Fix/uploader

eliawk:
- [x] a new user will not see the video list
- [x] show progress updates for upload to remote ipfs node

### DIFF
--- a/src/scripts/actions/UploaderActions.js
+++ b/src/scripts/actions/UploaderActions.js
@@ -18,8 +18,8 @@ import {
 } from 'constants/ActionConstants'
 import VideoRecord from 'records/VideoRecords'
 import { videoFetchSuccess } from 'actions/VideoActions'
-
 import type { Dispatch } from 'redux'
+import { store } from '../index'
 
 export const selectUploaderVideo = createAction(UPLOAD_VIDEO_SELECT)
 const uploadRequested = createAction(UPLOAD_REQUESTED)
@@ -32,6 +32,8 @@ const transcodingRequested = createAction(TRANSCODING_REQUESTED)
 const transcodingProgress = createAction(TRANSCODING_PROGRESS)
 const transcodingSuccess = createAction(TRANSCODING_SUCCESS)
 const transcodingFailure = createAction(TRANSCODING_FAILURE)
+
+console.log(store)
 
 // upload the video to the local ipfs node
 export const upload = (file: Object) => (dispatch: Dispatch<*>) => {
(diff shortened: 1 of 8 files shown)


src/scripts/actions/UploaderActions.js
```diff
@@ -33,6 +33,8 @@ const transcodingProgress = createAction(TRANSCODING_PROGRESS)
 const transcodingSuccess = createAction(TRANSCODING_SUCCESS)
 const transcodingFailure = createAction(TRANSCODING_FAILURE)
 
+console.log(store)
```
bent0b0x: What is this needed for? (if you are trying to inspect the redux store I _highly_ recommend using the `redux devtools` chrome extension

bent0b0x: https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en

src/scripts/index.js
```diff
@@ -10,7 +10,7 @@ import { getRoot } from 'utils/AppUtils'
 import createStore from 'scripts/createStore'
 import 'styles/app.scss'
 
-const store = createStore()
+export const store = createStore()
```
bent0b0x: Can we remove this? Shouldn't be a need in the app to export the store

src/scripts/selectors/UploaderSelectors.js
```diff
@@ -35,12 +35,26 @@ export const getSelectedUploaderVideo: (
   }
 )
 
-export const getUploadProgress = createSelector(
-  [getSelectedUploaderVideo],
-  upload => {
-    return (upload && upload.getIn(['uploadStatus', 'data', 'progress'])) || 0
-  }
-)
+// export const getUploadProgress = createSelector(
```
bent0b0x: What is wrong w/ these?

src/scripts/utils/AppUtils.js
```diff
@@ -51,8 +51,8 @@ export const getParatiiConfig = (env: ?string = 'development'): Object => {
 }
 
 export const prettyBytes = (num: ?number): string => {
-  if (num === null || num === undefined) {
-    num = 0
+  if (num === null || num === undefined || num === 0 || isNaN(num)) {
```
bent0b0x: Why check if `num === 0`? Wouldn't we want to print `0B` or something then?

Can this be simplified to `if (typeof num !== 'number' || isNaN(num))`?

PR #275, opened 2018-02-28
Uploader ui

jrgarou:
- Video and Share title responsive
- Fixed video list + video form responsive - they were too close to each to other.


src/scripts/components/Play.js
```diff
@@ -35,22 +35,43 @@ type State = {
 }
 
 const Wrapper = styled.div`
-  width: 100%;
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  flex: 0 0 100%;
+  margin: 0 auto;
```
bent0b0x: @jrgarou I just want to confirm, does this work well for both the `/play` and `/embed` routes?

jrgarou: @bent0b0x now it is working https://github.com/Paratii-Video/paratii-portal/pull/275/commits/6fb93d194ab2ae3e04fa599fe7cabdf40b3b5d6c

PR #293, opened 2018-03-02
Add first pass at front-end doc

bent0b0x:
@jellegerbrandy please let me know if you want me to cover anything else in here/if this should be formatted differently. I think this is a good first step but there are more things we can cover (perhaps in the future).


jellegerbrandy:
This is beautiful. Many thansk, Ben.

PR #301, opened 2018-03-02
Feature/refactor moda (=the staking modal) [do not merge yet]

jrgarou:
- Close modal by background click and ESC `keyCode`
- Call ModalStake by `string`. @bent0b0x 
For now, we pass a string that show the relative content. 
https://github.com/Paratii-Video/paratii-portal/blob/a4b96437703be6226c97319823778735207aeab3/src/scripts/components/VideoForm.js#L194
https://github.com/Paratii-Video/paratii-portal/blob/a4b96437703be6226c97319823778735207aeab3/src/scripts/components/widgets/modals/Modal.js#L93


jellegerbrandy:
Cool, thanks @jrgarou !

We need to solve this before we can merge it: https://github.com/Paratii-Video/paratii-portal/issues/304

src/scripts/components/VideoForm.js
```diff
@@ -186,16 +127,21 @@ class VideoForm extends Component<Props, Object> {
...
-    e.preventDefault()
-
-    this.props.showModal(
-      <ModalStake
-        videoId={this.state.id}
-        onSuccess={this.handlePublish}
-        user={this.props.user}
-      />
-    )
+  // onPublishSubmit (e: Object) {
```
bent0b0x: Delete this commented code?

src/scripts/components/VideoForm.js
```diff
@@ -204,7 +150,12 @@ class VideoForm extends Component<Props, Object> {
   }
 
   publishVideo (publish: false) {
-    this.saveData(publish)
+    this.props.openModal({
+      modal: 'ModalStake',
+      callback: () => {
```
bent0b0x: Can `callback` be named to be more descriptive? When is this called? Is it when the modal opens? Closes? Can this possibly be updated so that we have the option of passing in an `onOpen` and `onClose` callback?

src/scripts/components/VideoForm.js
```diff
@@ -234,67 +185,38 @@ class VideoForm extends Component<Props, Object> {
...
+            type="submit"
+            onClick={this.onPublishVideo}
+            disabled={!isPublishable}
+            purple
+          >
+            Publish
+          </Button>
+        </ButtonWrapper>
+      )
+    }
```
bent0b0x: ☝️ the way this is written totally works, but might want to look at [this conditional rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) pattern for the future as it is slightly more conventional.  [This one](https://reactjs.org/docs/conditional-rendering.html#inline-if-else-with-conditional-operator) is very common as well!

src/scripts/records/ModalRecord.js
```diff
@@ -0,0 +1,15 @@
...
+
+import { Record as ImmutableRecord } from 'immutable'
+
+class ModalRecord extends ImmutableRecord({
+  modalContent: '',
+  modalCallback: () => null,
+  showModal: false
+}) {
+  modalContent: string
+  modalCallback: () => null
```
bent0b0x: I don't think we want to enforce a type of `() => null`. Maybe just `() => any`? 

`() => void` if we do not want any return values

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -71,17 +73,26 @@ const Content = styled.div`
...
+      <Wrapper show={isVisible}>
+        <Container show={isVisible}>
           <CloseButton onClick={this.props.closeModal}>
             <SVG>
               <use xlinkHref="#icon-close" />
             </SVG>
           </CloseButton>
-          <Content>{this.props.children}</Content>
+          <Content>
+            {this.props.modalContent === 'ModalStake' && <ModalStake />}
```
bent0b0x: This is fine for now since we have just one modal (I think), but it might be good to move this to a `renderModal` method.

```javascript
renderModal() {
  const { modalContent } = this.props
  switch(modalContent) {
    case 'ModalStake':
      return <ModalStake />
  }
}
```

Then just call this method in `render`.

Also, I think it would be very good to not use raw strings for the modals. Instead, create a `constants/ModalConstants.js` file and define them as exportable constants

src/scripts/components/VideoForm.js
jrgarou: yes! deleted https://github.com/Paratii-Video/paratii-portal/pull/301/commits/ef56e9a93bb88ac25241e85ce446283d1bd98d44

src/scripts/records/ModalRecord.js
jrgarou: Done https://github.com/Paratii-Video/paratii-portal/pull/301/commits/ef56e9a93bb88ac25241e85ce446283d1bd98d44

src/scripts/components/VideoForm.js
jrgarou: @bent0b0x this _callback_ is more generic, used whenever it needs. In this case, it used when the submit is success on ModalStake

src/scripts/components/VideoForm.js
jrgarou: this code isn't changed from this PR and commits. 
Is this the code that you want to show?:

image

src/scripts/components/VideoForm.js
bent0b0x: Hm I'm not sure about that then. It is really unclear what `callback` does. Is this specific to `ModalStake`? If so shouldn't it be handled by `ModalStake` itself?

src/scripts/components/VideoForm.js
bent0b0x: It was just commented out before, I see now. Fine to leave the code as is, but FYI that there are more conventional patterns for conditional rendering

webpack.config.js
```diff
@@ -77,7 +77,8 @@ const config = {
   node: {
       net: 'empty',
       tls: 'empty',
-      dns: 'empty'
+      dns: 'empty',
+      fs: 'empty'
```
bent0b0x: Just for my understanding can you briefly explain why this change was needed? Thanks!

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -70,18 +72,32 @@ const Content = styled.div`
 `
 
 class Modal extends Component<Props, void> {
+  renderModal () {
```
bent0b0x: 🎉 nice

webpack.config.js
jrgarou: this is came from `dev` branch
https://github.com/Paratii-Video/paratii-portal/blame/dev/webpack.config.js#L81

bent0b0x: If it was still on `dev` it wouldn't be showing up in this diff 😸 

Maybe you need to re-merge `dev` again?

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -65,23 +81,58 @@ const SVG = styled.svg`
...
+    const { modalContent } = this.props
+    switch (modalContent) {
+      case 'ModalStake':
+        return <ModalStake />
+    }
+  }
   render () {
+    const isVisible = this.props.showModal
+    document.onkeydown = event => {
```
bent0b0x: This needs to be done in `componentDidMount`. Otherwise, every time this component renders we add this event listener.

Then in `componentWillUnmount` you should remove the event listener

src/scripts/components/widgets/modals/ModalStake.js
```diff
@@ -35,41 +40,48 @@ const Footer = styled.div`
...
-    // this.handlePublish = this.handlePublish.bind(this)
   }
 
   onSubmit (event: Object) {
     event.preventDefault()
 
     paratii.eth.tcr
       .checkEligiblityAndApply(
-        this.props.videoId.toString(),
+        this.props.videoId.id.toString(),
```
bent0b0x: Why is the prop called `videoId` if it has a property called `id`? Shouldn't the prop be called `video`?

src/scripts/components/widgets/modals/ModalStake.js
```diff
@@ -35,41 +40,48 @@ const Footer = styled.div`
...
+            description: this.state.description,
+            author: this.state.author,
+            published: true
+          }
+          this.props.saveVideoInfo(videoToSave)
+          this.props.closeModal()
           console.log(
-            `video ${this.props.videoId.toString()} successfully applied to TCR Listing`
+            `video ${this.props.videoId.id.toString()} successfully applied to TCR Listing`
           )
```
bent0b0x: All of this should really be in an `*Actions.js` module. We don't want this kind of complex logic in components. At the very least there should be a TODO comment added noting this

src/server/routes/embed.js
```diff
@@ -91,7 +91,7 @@ module.exports = async (req: $Request, res: $Response) => {
 
         <style>
           html {
-            font-size: 16px;
+            font-size: 14px;
```
bent0b0x: Was this change meant to be included? Does not seem relevant

src/scripts/components/widgets/modals/Modal.js
bent0b0x: And if you just set `.onkeydown` without using `document.on('keydown')` isn't this going to override any other keydown listeners we have set?

jrgarou: @bent0b0x Yeah! Its is override 
 I'll change this

src/server/routes/embed.js
jrgarou: This is not to be there. fixing that.

jrgarou: done https://github.com/Paratii-Video/paratii-portal/pull/301/commits/6baf7d4bf6bd521621ce67a64a8ea4bbee0f34e8

src/scripts/components/widgets/modals/ModalStake.js
jrgarou: Done https://github.com/Paratii-Video/paratii-portal/pull/301/commits/49f681e2a5dc7801ec9e0e4aa422af395946a54b

src/scripts/components/widgets/modals/Modal.js
jrgarou: https://github.com/Paratii-Video/paratii-portal/pull/301/commits/49f681e2a5dc7801ec9e0e4aa422af395946a54b and https://github.com/Paratii-Video/paratii-portal/pull/301/commits/bf833c6419be3b0e190bfc60e4e270d764f41157

src/scripts/components/widgets/modals/ModalStake.js
jrgarou: Let's move to a `*Action.js` ?

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -65,23 +75,57 @@ const SVG = styled.svg`
...
+  }
+`
+
 class Modal extends Component<Props, void> {
+  componentDidMount () {
+    document.addEventListener('keydown', event => {
+      if (event.keyCode === 27) {
+        this.props.closeModal()
+      }
+    })
```
bent0b0x: Looks much better. One more thing that should be done, as with the current way this is written we will add new listeners every time we mount the component and they will never be removed.

- move the callback into a method on this component
- then, define `componentWillUnmount`

```javascript
handleKeydown = (event) => {
  if (event.keyCode === 27) {
    this.props.closeModal()
  }
}

componentDidMount() {
  document.addEventListener(this.onKeydown);
}

componentDidUnmount() {
  document.removeEventListener('keydown', this.onKeydown);
}
```

jrgarou: Done https://github.com/Paratii-Video/paratii-portal/pull/301/commits/1afdcbafae637b40166bddfec94ae502b784ccbd

src/scripts/components/widgets/modals/ModalStake.js
```diff
@@ -91,60 +101,75 @@ class ModalStake extends Component<Props, Object> {
...
+    const selectedVideo = nextProps.selectedVideo
+    this.setState({
+      id: selectedVideo.id,
+      title: selectedVideo.title,
+      description: selectedVideo.description,
+      duration: selectedVideo.duration,
+      author: selectedVideo.author,
+      errorMessage: false,
+      agreedTOC: false // TODO,
+    })
```
bent0b0x: Just want to understand better. Why is all of this put in state when it can just be read from props?

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -65,23 +75,63 @@ const SVG = styled.svg`
...
+    if (event.keyCode === 27) {
+      this.props.closeModal()
+    }
+  }
+
+  componentDidMount () {
+    document.addEventListener('keydown', this.onKeydown)
+  }
+
+  componentDidUnmount () {
```
bent0b0x: Sorry I had a typo in my example, this is `componentWillUmnount`. There is no `componentDidUnmount`. My bad!

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -114,13 +114,12 @@ class Modal extends Component<Props, void> {
     document.addEventListener('keydown', this.onKeydown)
   }
 
-  componentDidUnmount () {
+  componentWillUnmount () {
```
bent0b0x: 🎉

jrgarou: 🕺🏿

src/scripts/components/widgets/modals/ModalStake.js
jrgarou: @ya7ya could we change that?

src/scripts/components/VideoForm.js
```diff
@@ -249,7 +234,7 @@ class VideoForm extends Component<Props, Object> {
         <VideoFormWrapper>
           <Form
             onSubmit={this.onPublishVideo}
-            disabled={this.props.selectedVideo.storageStatus.name === 'running'}
+            // disabled={this.props.selectedVideo.storageStatus.name === 'running'}
```
bent0b0x: Remove this if not needed please 😄

PR #311, opened 2018-03-05
Notification system

eliawk:
- success and error modal
- connected to redux status
- need some more style (@jrgarou)

// need `npm install` because it use two different packages


src/scripts/records/NotificationRecord.js
```diff
@@ -0,0 +1,25 @@
...
+    callback: null
+  }
+}) {
+  title: string
+  message: string
+  position: string
+  autoDismiss: number
+  action: {
+    label: '',
+    callback: () => void
```
bent0b0x: What is `callback`? I think it would be better to rename this to indicate when it is called. 

What about `onPerformAction`?

src/scripts/types/ApplicationTypes.js
```diff
@@ -32,7 +33,8 @@ export type RootState = {
   uploader: UploaderRecord,
   user: UserRecord,
   videos: VideoRecordMap,
-  player: PlayerRecord
+  player: PlayerRecord,
+  notifications: NotificationRecord
```
bent0b0x: Is this correct? Are we sure it is not `Array<NotificationRecord>`? Can there really be only one notification at once?

src/scripts/records/NotificationRecord.js
eliawk: Is a generic callback, this is just a default!

src/scripts/types/ApplicationTypes.js
eliawk: yes you are totally right, this is an `Array<NotificationRecord>`

src/scripts/records/NotificationRecord.js
bent0b0x: What is a generic callback? That is the problem, it is very unclear what this is supposed to do

PR #317, opened 2018-03-07
Fix/embed routing and404

geckoslair:


src/index.html
```diff
@@ -2,9 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <title>Paratii</title>
+    {{meta}}
```
bent0b0x: If we are calling this `meta` then should every `<meta>` tag be moved into a variable? We still have `<meta name="viewport">` below

src/server/app.js
```diff
@@ -1,10 +1,10 @@
 const express = require('express')
-const path = require('path')
+// const path = require('path')
```
bent0b0x: Can you remove this comment?

src/server/routes/index.js
```diff
@@ -0,0 +1,121 @@
...
+      if (video !== null) {
+        meta = openGraphHead(meta, video)
+        meta = twitterCardHead(meta, video)
+      } else {
+        meta = notFoundVideo(meta)
+      }
+
+      break
+    default:
+      console.log('default')
```
bent0b0x: Mind deleting this `console.log` (and any others)?

src/server/routes/index.js
```diff
@@ -0,0 +1,121 @@
...
+      meta = basicHead(meta)
+      script = '<script type="text/javascript" src="/bundle.js"></script>'
+      break
+  }
+
+  // console.log(index)
```
bent0b0x: Remove comment

src/server/routes/index.js
```diff
@@ -0,0 +1,121 @@
...
+
+  index = index.replace('{{meta}}', meta)
```
bent0b0x: Is this just a string replacement? I would rather us not do this. Can't we use [express-handlebars](https://github.com/ericf/express-handlebars) instead of fudging together our own template engine?

src/server/routes/index.js
```diff
@@ -0,0 +1,121 @@
...
+  index = index.replace('{{script}}', script)
+  console.log(meta)
+  console.log(script)
```
bent0b0x: Remove logging

src/server/routes/index.js
```diff
@@ -0,0 +1,121 @@
...
+  meta += `<meta property="og:video:secure_url" content="${embedUrl}">`
+  meta += `<meta property="og:video:type" content="text/html">`
+  meta += `<meta property="og:video:width" content="${width}" />`
+  meta += `<meta property="og:video:height" content="${height}" />`
+  meta += `<meta property="og:type" content="video.other" />`
+  meta += `<meta property="og:url" content="${embedUrl}" />`
+  meta += `<meta property="og:title" content="${video.title}" />`
+  meta += `<meta property="og:image" content="${thumbnailUrl}" />`
+  meta += `<meta property="og:description" content="${video.description}" />`
+
```
bent0b0x: I think it would be best if we use a template engine and move all of this into a template partial, rather than a large string concatenation

src/server/app.js
```diff
@@ -28,12 +28,15 @@ if (process.env.NODE_ENV === 'production-notugly') {
   )
 }
 
-app.use(express.static(path.resolve(__dirname, '../../', 'build')))
-app.get('/embed/:id', videoRoute)
-app.get('/play/:id', videoRoute)
+// app.use(express.static(path.resolve(__dirname, '../../', 'build')))
```
bent0b0x: Why is this removed? I think this would break a lot

src/server/app.js
```diff
@@ -28,12 +28,15 @@ if (process.env.NODE_ENV === 'production-notugly') {
...
+app.get('/embed/:id', routeHelper)
+app.get('/', routeHelper)
 app.get('/oembed', oembedRoute)
-app.get('*', (req, res) => {
-  res.sendFile(path.resolve(__dirname, '../../', 'build', 'index.html'))
-})
+app.get('*', routeHelper)
+// app.get('*', (req, res) => {
+//   res.sendFile(path.resolve(__dirname, '../../', 'build', 'index.html'))
+// })
```
bent0b0x: I would think this comment would break loading the app when you visit `localhost:8080/upload` directly

bent0b0x:
@geckoslair I think we need to use a formal templating engine instead of string concatenation. Additionally, I think there are changes to our route handler that need to be tested before merging this.

bent0b0x:
https://github.com/ericf/express-handlebars

src/server/app.js
geckoslair: i've removed because serves also index.html, do you know how to fix that?

src/server/app.js
geckoslair: no, it works.

src/server/routes/index.js
geckoslair: sound better

src/server/routes/index.js
geckoslair: done

src/server/routes/index.js
geckoslair: ok

src/index.html
geckoslair: ok

src/server/app.js
geckoslair: ok

src/server/routes/index.js
geckoslair: ok

src/server/routes/index.js
geckoslair: ok

src/server/app.js
bent0b0x: Without this I would not expect any of our asset loading to work (i.e. images will not load).

I'm sure there is an `express` api for excluding certain files. However, why do we need to exclude `/index.html`?

src/server/app.js
bent0b0x: Ah, perhaps that is because `routeHelper` is doing the same thing?

src/server/app.js
geckoslair: because it wouldn't get in the get request. I'll check with the handlebar features.

PR #323, opened 2018-03-08
Updating documentation

LorenzoAlluminio:
Added a few troubleshooting tips


bent0b0x:
Looks good! The only thing I would say is that installing `nodemon` globally should not be necessary? Can you clarify why this might be needed? I do not think it should be

LorenzoAlluminio:
I've ran the two commands in sequence so actually I'm not sure if the last one is enough, but maybe you're right :)
let's hear @geckoslair , he was guiding me through the fixes :)

bent0b0x:
Got it. Feel free to chime in @geckoslair but when our `server:dev` script runs we are using the `nodemon` installed in `node_modules`, not the global one. So I am fairly confident this step can be removed :).

As for the other part of the port fix, can you lmk where that came from just so I can understand the problem better?

Thanks for this documentation!

LorenzoAlluminio:
Ok I'll remove it :+1: 

The fix comes from [here](https://stackoverflow.com/questions/34662574/node-js-getting-error-nodemon-internal-watch-failed-watch-enospc).
Honestly I don't know what it does, but for me it did the trick

bent0b0x:
Thanks for the context. Couldn't hurt to link to the stack overflow answer in the readme, there is a bit of explanation as to the "why this works" in there. Once the `npm i -g nodemon` is removed I'd say you are good to merge this!

LorenzoAlluminio:
Now it should be fine, added a few more graphic improvements as well.
Thank you :smile:

README.md
```diff
@@ -127,3 +127,24 @@ If you get the following error:
...
+If you get the following error:
+
+  ```
+  $ [nodemon] Internal watch failed: watch .../paratii-portal/node_modules ENOSPC
+  ```
+Run the following code to fix the port issue:
+
+    $ echo fs.inotify.max_user_watches=582222 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
+Explaination [here](https://stackoverflow.com/questions/34662574/node-js-getting-error-nodemon-internal-watch-failed-watch-enospc)
```
bent0b0x: 🎉

PR #324, opened 2018-03-08
UI fixes and updates

jrgarou:
- Moved styles from index.html to `app.scss` @bent0b0x is this fine for embed? It looks ok.
- Wrap the **video** and the **descritption Card**(no info yet)  on `/play` - the `@media` rules gets both them
- Play icon on `VideoFormInfoBox`
image
- `VideoListItem` with @pedrocasa design: thumb, truncate title, status and a small bar(`<VideoProgressBar ... small />`. 
image

- `VideoListItem.onClick` Scroll to `VideoForm` when `VideoForm.offsetTop > window.innerHeight / 2` .

- Updated header and footer size.
- Unbold the headers(`<h1>,<h2>...`)


src/scripts/components/Play.js
```diff
@@ -540,86 +565,93 @@ class Play extends Component<Props, State> {
     } else {
       return (
         <Wrapper isEmbed={isEmbed}>
-          <PlayerWrapper
```
bent0b0x: This big diff is kind of scary, but if the new player works it works!

Is this mostly because the player is now wrapped in `VideoWrapper`?

src/styles/app.scss
```diff
@@ -14,4 +14,30 @@ html, body, #root {
...
+  height: 50px;
+  left: 50%;
+  position: fixed;
+  top: 50%;
+  transform: translate3d(-50%, -50%, 0);
+  width: 50px;
+}
+
+h1, h2, h3, h4, h5, h6 {
+  font-weight: 400;
```
bent0b0x: I don't think this is going to work for embed. `embed/index.js` imports a separate `/embed/index.scss` file, not this one

src/scripts/components/VideoListItem.js
```diff
@@ -76,6 +107,13 @@ class VideoListItem extends Component<Props, void> {
   }
 
   handleClick () {
+    const form = document.getElementById('video-form')
+    const offsetTop = form.offsetTop - 20
+    if (form) {
+      if (offsetTop > window.innerHeight / 2) {
+        document.getElementById('root').scrollTo(0, offsetTop)
+      }
+    }
```
bent0b0x: This works, but there is a more react-way to get references to actual dom nodes 😄 (requires no hard-coded IDs).

Take a look at [this line](https://github.com/Paratii-Video/paratii-portal/blob/460fa75750281b9b66a0ad93a4d92dd3e9c72066/src/scripts/components/widgets/VolumeBar.js#L107) in `VolumeBar.js`. Here, we take advantage of `styled-components`'s [innerRef](https://github.com/styled-components/styled-components/blob/master/docs/tips-and-tricks.md#refs-to-dom-nodes) property. This provides a mechanism for us to access the [ref property](https://reactjs.org/docs/refs-and-the-dom.html) on the underlying react elements.

Long story short: `ref/innerRef` are properties that take a callback. This callback is invoked with an `HTMLElement` that you can then cache/store somewhere. `VolumeBar.js` is a good example.

Let me know if you have any questions about how this should work!

src/scripts/components/VideoListItem.js
```diff
@@ -76,6 +107,14 @@ class VideoListItem extends Component<Props, void> {
...
 
   handleClick () {
+    const form = document.getElementById('video-form')
+    if (form) {
+      const offsetTop = form.offsetTop - 20
+
+      if (offsetTop > window.innerHeight / 2) {
+        document.getElementById('root').scrollTo(0, offsetTop)
+      }
+    }
```
bent0b0x: This works, but there is a more react-way to get references to actual dom nodes 😄 (requires no hard-coded IDs).

Take a look at this line in VolumeBar.js. Here, we take advantage of styled-components's innerRef property. This provides a mechanism for us to access the ref property on the underlying react elements.

Long story short: ref/innerRef are properties that take a callback. This callback is invoked with an HTMLElement that you can then cache/store somewhere. VolumeBar.js is a good example.

Let me know if you have any questions about how this should work!

src/styles/app.scss
jrgarou: @bent0b0x, could we import all styles from `src/styles/app.scss`?

src/scripts/components/Play.js
jrgarou: @bent0b0x yes. Basically it's just about the wrapper from `/play` and `/embed`

src/scripts/components/Play.js
```diff
@@ -540,86 +565,93 @@ class Play extends Component<Props, State> {
...
+                          requestFullscreen(this.wrapperRef)
+                        } else {
+                          requestCancelFullscreen()
+                        }
+                      }}
+                    />
+                  </OverlayWrapper>
+                )}
+              </Transition>
+              <Player id="player" />
```
eliawk: this `player id="player"` is still there hardcoded? is it correct @bent0b0x?

src/scripts/components/VideoListItem.js
jrgarou: I changed this code in https://github.com/Paratii-Video/paratii-portal/pull/324/commits/acceeb58b913e65813af3ed20a546c5fe1a4aff0. I'm trying to get the reference of `VideoForm` on `VideoListItem` from `VideoManagerContainer`.

When I [click on](https://github.com/Paratii-Video/paratii-portal/blob/acceeb58b913e65813af3ed20a546c5fe1a4aff0/src/scripts/components/VideoListItem.js#L109-L112) a `VideoListItem` the `this.props.videoFormRef` returns `null`. After somes clicks it get the `div.Card`.
if you click on a `VideoListItem` then click on `ADD ANOTHER VIDEO` then again clicks on `VideoListItem` and keep clicking on it(no change to other `VideoListItem`) the `this.props.videoFormRef` will always returns `null`:
getref

src/scripts/components/Play.js
bent0b0x: This is needed because of `clappr`. I don't think we have a feasible way around this right now

src/styles/app.scss
bent0b0x: You can update `embed/index.js`. Just note that we want to keep `/embed` light so if there things in here that the embed doesn't need let's not import them. Otherwise, import what is needed into `/embed/index.scss`

src/scripts/components/VideoListItem.js
```diff
@@ -76,6 +107,7 @@ class VideoListItem extends Component<Props, void> {
   }
 
   handleClick () {
+    console.log(this.props)
```
bent0b0x: remove log

src/styles/app.scss
jrgarou: Done https://github.com/Paratii-Video/paratii-portal/pull/324/commits/2800e1ddba828514e8a922e32a7618f0b3acab8e

src/scripts/components/VideoListItem.js
jrgarou: done https://github.com/Paratii-Video/paratii-portal/pull/324/commits/fd247c6ae5a0a1a4e0a5e563e6638cd8c96f3dd6

jellegerbrandy:
@jrgarou, could you fix the conflicts, so I can merge this? (go to your branch and do `git merge dev`)

PR #328, opened 2018-03-08
change the configuration of test for use a new chimp configuration

Leeaandrob:
Creating a new chimp configuration for use on dev and circleci


bent0b0x:
I’m not a huge fan of having two configs with a lot of duplicate info. Instead, can we have one config extend the other? Or try overriding one config via CLI args? If we have two karma configs I think we should do the same there. 

If we don’t do this the two will definitely get out of sync

jellegerbrandy:
I agree with @bent0b0x 

Let's do the extend thing.

So w'e have one default file called `chimp.config.js` that is used for development.

And another `chimp.circleci.config.js` that looks like this:

    config = require('chimp.config.js')
    config[webdriverio] = different config
    default export config

We can also move the `circleci` files in the `.circleci` directory, so it lookes less messy.

.circleci/karma.circle.conf.js
```diff
@@ -0,0 +1,16 @@
...
+        base: 'ChromeHeadless',
+        flags: ['--no-sandbox']
+      }
+    }
+  })
+}
+
+const config = Object.assign(changes, require('../karma.conf.js'))
+
+module.exports = config
```
bent0b0x: There are still karma changes here

PR #330, opened 2018-03-08
Feature/new chimp

Leeaandrob:
Finish the configurations of chimp only.


bent0b0x:
Still karma changes in here 😄

Leeaandrob:
I removed the .circle/karma.circleci.conf.js @bent0b0x

bent0b0x:
there is still a file rename in here related to `karma`. Check the [diff](https://github.com/Paratii-Video/paratii-portal/pull/330/files)

jellegerbrandy:
We must do these changes both for karma and for chimp, same pattern, we can do this in a single PR. Or am I missing a point?

bent0b0x:
I suggested just doing karma in another PR (as long as @Leeaandrob creates an issue for it) just because karma's is _slightly_ more complex. But we can do it in the same PR.

@Leeaandrob basically you need to take the objects that are passed to the karma functions and abstract those into their own modules that you can override and extend

Leeaandrob:
@bent0b0x I will follow the @jellegerbrandy desire using the same PR.

.circleci/karma.circleci.conf.js
```diff
@@ -0,0 +1,19 @@
+const shared = require('../karma.conf.js')
+
+module.exports = function(karma) {
+  shared(karma)
```
bent0b0x: I'm not sure extending it this way will work. Is this still work in progress?

Leeaandrob: https://circleci.com/gh/Paratii-Video/paratii-portal/1291 @bent0b0x

Issue #325, opened 2018-03-08
Video error on safari mobile


eliawk:
- go to embed page https://portal.paratii.video/embed/65te9Z7bXDM4
- you get this promise error
`Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'this.container.isPlaying')`

bent0b0x:
Resolved via #338

PR #338, opened 2018-03-10
New controls fix

bent0b0x:
**Issue**
#325 

The issue was a runtime error in clappr because they did not perform a `null` check (@jellegerbrandy this is why flow is great and they should use it 😄).

**Solution**
Upgrade clappr; they fixed the issue in [this commit](https://github.com/clappr/clappr/commit/d8270d4428337ea2084f21c5e2879347837a3a1e).


bent0b0x:
@geckoslair thanks for fixing the dependency :)

PR #339, opened 2018-03-11
Fix embed route

bent0b0x:
**Problem**
- `/embed/:id` was loading the regular portal, not the specialized embed route. And once it got there you always saw `404`

**Work Done**
- fix how embed param was passed to template
- fix embed styles processing

@geckoslair could you do some testing to make sure this is all good? My fix works at a high level but if you can make sure all of the partials are rendering as expected that would be great 😄

### DIFF
--- a/src/server/routes/index.js
+++ b/src/server/routes/index.js
@@ -19,11 +19,11 @@ exports.player = async function player (req, res, next) {
   const { id } = req.params
   const path = req.route.path
   const video = await paratii.core.vids.get(id)
-  const meta = ''
 
   res.render('index', {
     player: true,
     video: video,
+    embed: path === '/embed/:id',
     helpers: {
       title: function () {
         if (video !== null) {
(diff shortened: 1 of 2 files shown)


src/server/routes/index.js
```diff
@@ -19,11 +19,11 @@ exports.player = async function player (req, res, next) {
   const { id } = req.params
   const path = req.route.path
   const video = await paratii.core.vids.get(id)
-  const meta = ''
```
bent0b0x: This was not used AFAICT

src/server/routes/index.js
```diff
@@ -53,19 +53,6 @@ exports.player = async function player (req, res, next) {
...
-        return meta
-      },
-      embed: function () {
-        return path === '/embed/:id'
-      },
-      script: function () {
-        if (path === '/embed/:id') {
-          return '<script type="text/javascript" src="/embed/bundle.js"></script>'
-        } else if (path === '/video/:id') {
-          return '<script type="text/javascript" src="/bundle.js"></script>'
```
bent0b0x: This was not used AFAICT

src/server/routes/index.js
```diff
@@ -53,19 +53,6 @@ exports.player = async function player (req, res, next) {
       },
       embedUrl: function () {
         return 'https://portal.paratii.video/embed/' + video._id
-      },
-      meta: function () {
-        return meta
-      },
-      embed: function () {
-        return path === '/embed/:id'
```
bent0b0x: This was the culprit. We needed to pass this directly as a boolean, not as a render helper

webpack.config.js
```diff
@@ -102,16 +102,7 @@ const config = {
       {
         test: /\.scss$/,
         include: stylesDir,
-        exclude: stylesDir + "/embed/index.scss",
```
bent0b0x: Now we just inject the styles into the page w/ javascript like all of our other styles

Issue #329, opened 2018-03-08
Safari - player controls are flashing


eliawk:
![mar-08-2018 14-46-08](https://user-images.githubusercontent.com/362016/37167052-8f113976-22df-11e8-9ff8-d7c0908f6c6c.gif)

Issue #326, opened 2018-03-08
Play pause button


eliawk:
If you click on play the button disappear for some secs and and the come back with the pause icon.
Maybe we could you a sprite image or an svg morphing?

PR #346, opened 2018-03-13
326 - Play/pause button flicker fix

bent0b0x:
**Issue**
#326 

**Problem**
Our SVGs were not loading quickly enough to prevent a flicker in the UI when switching icons on buttons.

**Work Done**
Load SVGs in player controls (play, pause, volume, mute, fullscreen) inline so they do not have to be fetched from the server at runtime.

**Demo**

playpausewin

**Note**
I think a better way to solve this in the future would be to implement better cacheing for our assets. However, I think we would have to change our build process to name our static assets more dynamically so as to avoid allowing the cache to get stale. For now, I think this is the quickest and best solution.


webpack.config.js
```diff
@@ -1,140 +1,158 @@
...
             }
           }
         ]
       },
+      {
+        test: /\.(svg)$/,
+        include: assetsDir,
+        use: [
+          {
+            loader: 'svg-url-loader'
```
bent0b0x: This file got prettified, but this is the only important part of the PR

bent0b0x:
This also resolves #329. It was just much worse in Safari :'(

jellegerbrandy:
If you fix the conflicts, afaic you can merge it as well :-)

Issue #194, opened 2018-02-20
reduce bundle.js size


ya7ya:
currently the production build is 8.5MB which is too big, I used webpack-bundle-size-analyzer here are the results.

```bash
paratii-mediaplayer: 4.54 MB (32.3%)
clappr: 1.1 MB (7.84%)
paratii-contracts: 871.12 KB (6.05%)
node-forge: 813.28 KB (5.64%)
lodash: 567.3 KB (3.94%)
bitcore-lib: 327.09 KB (2.27%)
async: 248.23 KB (1.72%)
web3-eth-accounts: 211.74 KB (1.47%)
  bluebird: 167.25 KB (79.0%)
  eth-lib: 23.43 KB (11.1%)
  uuid: 5.83 KB (2.75%)
  <self>: 15.23 KB (7.19%)
bluebird: 174.97 KB (1.21%)
web3-core-promievent: 169.34 KB (1.18%)
  bluebird: 167.25 KB (98.8%)
  <self>: 2.09 KB (1.23%)
unorm: 139.82 KB (0.970%)
immutable: 139.01 KB (0.965%)
web3-eth-abi: 121.56 KB (0.844%)
  bn.js: 85.14 KB (70.0%)
  <self>: 36.41 KB (30.0%)
bitcore-mnemonic: 118.28 KB (0.821%)
web3-utils: 112.79 KB (0.783%)
  bn.js: 85.14 KB (75.5%)
  <self>: 27.64 KB (24.5%)
elliptic: 108.94 KB (0.756%)
bignumber.js: 97.65 KB (0.678%)
react-dom: 93.92 KB (0.652%)
  object-assign: 2.06 KB (2.19%)
  <self>: 91.86 KB (97.8%)
simple-peer: 91.74 KB (0.637%)
  readable-stream: 67.81 KB (73.9%)
    inherits: 672 B (0.968%)
    <self>: 67.16 KB (99.0%)
  process-nextick-args: 1.02 KB (1.11%)
  <self>: 22.9 KB (25.0%)
ethjs-unit: 90.22 KB (0.626%)
  bn.js: 85.14 KB (94.4%)
  <self>: 5.08 KB (5.63%)
number-to-bn: 86.96 KB (0.604%)
  bn.js: 85.14 KB (97.9%)
......
```

As you can see, we should probably work on optimizing the `paratii-mediaplayer` first since it's the biggest dependency.

ya7ya:
[...]

PR #348, opened 2018-03-14
GZIP!! and minifying all but IPFS

ya7ya:
this is part of the effort to properly uglify the production build. 
I managed to split IPFS in a separate bundle here https://github.com/Paratii-Video/paratii-lib/pull/162
but this still breaks in `production` mode. i'm debugging the issue 😞 

ref https://github.com/Paratii-Video/paratii-portal/issues/194#issuecomment-372827913


bent0b0x:
@ya7ya if we follow the dependency philosophy that our libraries should not come w/ any dependencies and instead rely on applications to provide them, then I think our problem becomes easier. Make `ipfs` a `peerDependency` and then tell `webpack` in `paratii-lib` to not bundle it and instead rely on it to be provided later. See [example in paratii-mediaplayer](https://github.com/Paratii-Video/paratii-mediaplayer/blob/master/webpack.config.js#L87-L92). `paratii-mediaplayer` does not come w/ any dependencies bundled. 

Then, in the portal we can easily update the `webpack` config to treat `ipfs-js` differently.

So moving to `peerDependencies` helps in this case but I think it is also a good idea because it prevents double dependency bundling.

jellegerbrandy:
Ok, there is one part i do not understand yet about peerdependencies. No,
two :-)

1. As far as I can see, we are not using a "bundled" version of paratii-lib
for the portal bunble (
https://github.com/Paratii-Video/paratii-portal/blob/dev/src/scripts/utils/ParatiiLib.js#L3).
So it is unclear to me why peerdepdenceis would make a difference here.

2. More of a general question: we depend on 10s of packages and 100s of
modules, all not using peer dependencies. It is unclear to me why
paratii-lib in particular is problematic here and needs to use peer
dependencies..

bent0b0x:
1. If we are not using the bundled version, then all of the dependencies are already being treating as peer dependencies. Meaning that the portal has to install and provide them all. If this is the recommended way to use `paratii-lib`, then I think `paratii-lib` should mark all of the deps as peer deps so it is clear that any use of the lib needs to install these things or else the library will not work.

So, if `paratii-portal` is already the one providing `ipfs-js`, @ya7ya why can't we just update portal's webpack.config.js to not mangle or uglify `ipfs-js`? Or mangle it specially?

2. I could be wrong, but I think it is just not standard practice b/c it makes libraries more difficult to install and use when you use peer dependencies. But in the end it definitely keeps your bundle size smaller and prevents duplicate dependencies if the app is the only one providing dependencies.

ya7ya:
I just realized that the express server here didn't have the `compression` middleware!!! , this significantly reduces the size of the files to under 1mb 💨

PR #349, opened 2018-03-14
47 - Playback Quality

bent0b0x:
**Issue**
[paratii-mediaplayer#47](https://github.com/Paratii-Video/paratii-mediaplayer/issues/47)

**Work Done**
Implement a video quality plugin for our controls.

**Demo**

qualitywin

***

**TODO**
- [x] determine correct initial quality level (our guess is not always right)
- [x] figure out bug where slide modal animation becomes extremely choppy
- [x] improve UX when switching qualities. sometimes it hangs w/ no loading indicator

cc @pedrocasa let me know what you think. This is much closer to your design now but please feel free to point out anything that needs to be changed.


src/scripts/components/Play.js
```diff
@@ -223,10 +233,25 @@ class Play extends Component<Props, State> {
...
       })
     }
+
+    const playerNode = document.querySelector(`#${PLAYER_ID}`)
+    if (playerNode) {
+      const parentNode: ?Node = playerNode.parentNode
+      if (parentNode) {
+        parentNode.removeChild(playerNode)
+      }
+    }
```
bent0b0x: ☝️ this is a safeguard to try and finally remove the duplicate player problems. In case `clappr` hasn't cleaned up our `#player` node, we will.

bent0b0x:
@jellegerbrandy I think this is good for a test from you and @pedrocasa. It is not perfect but all of the basics work. Still some room for UX improvement.

bent0b0x:
image

☝️ this branch is mostly in line w/ the mock above but there are a few tweaks to be made

src/scripts/components/Play.js
bent0b0x: Yes, you are right. I think moving this to `componentWillUnmount` is probably a more sane for it to be.

bent0b0x: Changed via e7212e3

PR #356, opened 2018-03-14
removed a commented but useful part

eliawk:


jellegerbrandy:
I'll merge this, with the understanding that this is a temporary fix while waiting for https://github.com/Paratii-Video/paratii-portal/issues/357

Issue #67, opened 2018-01-22
Implement design for player page in portal


jellegerbrandy:

pedrocasa:
first sketch of the video page

![00](https://user-images.githubusercontent.com/30123644/35438834-439d6a76-027f-11e8-82cc-122791f30b6e.png)

felipegaucho:
Closing this, although the **user icon** must be taken away from the embedded player in this context (portal).

pedrocasa:
Done in this image @felipegaucho

jellegerbrandy:
We have this design now, we can create separate issues for particular features or problems

PR #361, opened 2018-03-15
67 player page

jrgarou:
### Implement design for player page in portal #67
https://github.com/paratii-video/paratii-portal/issues/67

- The `this.props.video.free` always return false.
- The **video**, **like** and **dislike** buttons aren't working.

image


src/scripts/components/Play.js
```diff
@@ -676,7 +698,46 @@ class Play extends Component<Props, State> {
...
+                    <Text small gray>
+                      0
+                    </Text>
+                  </ButtonIcon>
+                  <ButtonIcon>
+                    <SVG>
+                      <use xlinkHref="#icon-play-like" />
+                    </SVG>
+                    <Text small gray>
+                      0
```
bent0b0x: Do we want to hard code in these values?

src/styles/app.scss
```diff
@@ -4,11 +4,6 @@ html, body, #root {
   width: 100%;
 }
 
-#root {
-  overflow-x: hidden;
-  overflow-y: scroll;
```
bent0b0x: Same question re: scrolling as in other PR

package-lock.json
```diff
@@ -3565,6 +3565,7 @@
           "requires": {
             "anymatch": "1.3.2",
             "async-each": "1.0.1",
+            "fsevents": "1.1.3",
```
bent0b0x: Can you reset `package-lock.json` on this branch to be what is on `dev`? Just because no deps were actually updated.

src/styles/app.scss
eliawk: i don't get you point @bent0b0x, we need this overflow fix?

src/scripts/components/Play.js
jrgarou: I was thinking to create a component for our SVGs. But for now, let's keep like this.
What you think, @bent0b0x?

bent0b0x: I think this is fine for now. In my comment above though I was talking about the `0`s that are hard coded

src/styles/app.scss
bent0b0x: We did at some point, im just making sure we’ve tested and made sure that it is no longer necessary

jrgarou: ...and it could causes a "double scroll"
image

The header its above the scroll because of the `position: fixed`

Issue #359, opened 2018-03-15
Sharing overlay opacity settings (make more transparent)


felipegaucho:
A minimal thing. The "sharing" overlay is 100% opaque as of today:

<img width="969" alt="screen shot 2018-03-15 at 11 53 16 am" src="https://user-images.githubusercontent.com/26517120/37482628-79e04da6-2863-11e8-8016-61df4f367168.png">

References are not 100% opaque, usually there's some transparency to see the video behind (we were like this, until some days ago):

<img width="436" alt="screen shot 2018-03-15 at 3 14 52 pm" src="https://user-images.githubusercontent.com/26517120/37482786-ea5d368e-2863-11e8-9a0b-9c41d169182e.png">

@pedrocasa do we have specs on this?

jellegerbrandy:
It is still completely opaque in production, @jrgarou

jellegerbrandy:
apologies. The problem was #392

PR #362, opened 2018-03-15
Description height - fixed #354  - Sharing overlay opacity - solved #359

### LINKED ISSUE #354: Description height
Step to reproduce:
- make a very long description
- save data
- refresh page
- you go back to the video and you see this
 
schermata 2018-03-14 alle 12 17 25

jrgarou:
- Description height
https://github.com/paratii-video/paratii-portal/issues/354
Handling the text area height when `componentDidMount`

- Sharing overlay opacity settings (make more transparent) #359
https://github.com/Paratii-Video/paratii-portal/issues/359

@pedrocasa I've set the opacity to `0.9`

### DIFF
--- a/src/styles/app.scss
+++ b/src/styles/app.scss
@@ -4,11 +4,6 @@ html, body, #root {
   width: 100%;
 }
 
-#root {
-  overflow-x: hidden;
-  overflow-y: scroll;
-}
-
 .full-block {
   display: block;
   height: 100%;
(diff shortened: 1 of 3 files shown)


src/styles/app.scss
```diff
@@ -4,11 +4,6 @@ html, body, #root {
   width: 100%;
 }
 
-#root {
-  overflow-x: hidden;
-  overflow-y: scroll;
-}
```
bent0b0x: Just want  to make sure, we are still able to scroll on small screen heights? Or when the user has zoomed in?

jrgarou: @bent0b0x we are using the default scroll.
I don't get it the point

Issue #366, opened 2018-03-16
Voucher code succeeds, notifications say I'm good - but 0 PTIs in the wallet!


felipegaucho:
It all goes right, but the counter remains at "0".

![giphy 10](https://user-images.githubusercontent.com/26517120/37547118-8e556db6-294e-11e8-81a6-f80cc3401f17.gif)

**NOTE: When I refresh the page, the missing PTI are miraculously there.**

bent0b0x:
@felipegaucho let me take a look at this, the fix should be simple

bent0b0x:
And as an FYI, we have basic repetitive polling logic that refreshes the balance every 3 minutes (it used to be like every 5 seconds). Maybe we can find a happy medium.

For this issue, though, I will refresh the balance once staking is successful

felipegaucho:
> And as an FYI, we have basic repetitive polling logic that refreshes the balance every 3 minutes (it used to be like every 5 seconds). Maybe we can find a happy medium.

That might be it. I'm sure @jellegerbrandy and the guys are in favour of as minimum polling as possible, but now I published a video, made a stake, and the balance didn't update until I noticed much later. Can we refresh it once a stake is made, too?

bent0b0x:
@felipegaucho we certainly can. Obviously this kind of solution is not sustainable so we need to come up w/ a plan to either poll or maintain a socket connection to always live update this value. 

Before I implement this, are there other functionalities in our portal right now that you can think of that update your PTI balance?

felipegaucho:
As we spoke today, so far just redeem voucher and publish+stake. Happy to close this issue once https://github.com/Paratii-Video/paratii-portal/pull/370 is working fine

bent0b0x:
@felipegaucho sounds good. I just merged that PR but we can leave this open until it is in Production

PR #367, opened 2018-03-16
366 - refresh token balances after voucher redemption

bent0b0x:
**Issue**
#366 

**Problem**
After redeeming PTI via a voucher, the balance in the nav bar does not update after either 3 minutes have passed or the user reload the page. This is not ideal!

**Work Done**
Trigger a balance update once the voucher has been redeemed. 

**Note**
~~I do not have a voucher code to test this with so if someone could send me one that would be great 😄~~
☝️ I got some vouchers, tested, and it works

### DIFF
--- a/src/scripts/components/widgets/RedeemVoucher.js
+++ b/src/scripts/components/widgets/RedeemVoucher.js
@@ -8,6 +8,7 @@ import Text from '../foundations/Text'
 import Card from '../structures/Card'
 
 type Props = {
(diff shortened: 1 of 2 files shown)


bent0b0x:
@jellegerbrandy do you think it is possible to write a chimp test for this? Can we generate voucher codes that will always work in the test environment?

bent0b0x:
I am going to merge this as it is a very small, simple PR for a bug fix

PR #368, opened 2018-03-17
Prevent super tall video manager

bent0b0x:
**Problem** 
When you have uploaded a lot of videos, your video list can grow super tall making the other activities on the upload page too tall to be usable.

heightfail

***

**Fix**
Set a max height for the video manager and add scrolling to the video list

heightwin

### DIFF
--- a/src/scripts/components/VideoList.js
+++ b/src/scripts/components/VideoList.js
@@ -27,6 +27,7 @@ const List = styled.ul`
   display: block;
   padding-bottom: 80px;
   width: 100%;
+  overflow: scroll;
 `
 
 class VideoList extends Component<Props, void> {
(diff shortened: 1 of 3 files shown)


jrgarou:
@bent0b0x this is the VideoList scroll:
image

Perhaps we need to use a stylized scroll. And using javascript for that. Perhaps the better solution for that is using a component to create a stylized scroll.
Do you know a _well know/used_ stylized scroll react plugin/component?


But @pedrocasa is restructuring this page, and the VideoList should be out.

jrgarou:
@bent0b0x I think just CSS will not resolve. To get the same results, whatever device/browser/etc, I think we should use some react component.

And the `max height` should be on `VideoList` to avoid some problems:


image

bent0b0x:
I think you are right @jrgarou. I can do a little research into custom scroll bar react components that could integrate seemlessly into our app!

jrgarou:
And perhaps just set a `max-height` just for `VideoList`. 
But I think the `video list` doesn't work well in this context. When it goes to the tablet breakpoint, the `video list` goes to the top and its "navigation" could be annoying. That's why I was trying to do that "scrollTo(`react.ref`)" `VideoForm`. 

But @pedrocasa is designing a better solution for that
image

Issue #360, opened 2018-03-15
Progress bar on player never disappears


felipegaucho:
Even when the mouse goes away, it remains there, on the bottom. No controls should remain once the video is playing or buffering and the 'mouse' is away,

![giphy 9](https://user-images.githubusercontent.com/26517120/37483448-cced4e8a-2864-11e8-8615-be66fb85bd37.gif)

jrgarou:
Looks fine to me

jellegerbrandy:
yeah, sorry, looks fine to me as well. I had checked prodcution but there was an issue with thecontinuous deployment so it was an old version.

PR #369, opened 2018-03-17
360 - fully hide controls and fix player bug

bent0b0x:
**Issue**
#360 

**Problem**
The seek bar never fully hid when playing a video

**Work Done**
- fully hide controls
- when playing/pausing via space bar, prevent browser from also scrolling on this key press (it led to a very bad UX)

**Demo**

hidecontrolswin

cc @felipegaucho

### DIFF
--- a/src/scripts/components/Play.js
+++ b/src/scripts/components/Play.js
@@ -426,6 +426,7 @@ class Play extends Component<Props, State> {
 
     // Space key
     if (e.keyCode === 32) {
+      e.preventDefault()
       this.togglePlayPause()
     }
   }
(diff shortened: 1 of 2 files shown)


src/scripts/components/Play.js
```diff
@@ -426,6 +426,7 @@ class Play extends Component<Props, State> {
 
     // Space key
     if (e.keyCode === 32) {
+      e.preventDefault()
```
bent0b0x: This prevents scrolling when the user presses "space" in an attempt to play/pause the video

Issue #365, opened 2018-03-16
Video is ready - color and text


eliawk:
Change color just after save changes.

<img width="449" alt="schermata 2018-03-16 alle 17 54 15" src="https://user-images.githubusercontent.com/362016/37544245-0e4f3bc0-2943-11e8-8dde-4a4178303a98.png">

eliawk:
```
if (
      video.storageStatus.name === 'success' &&
      video.transcodingStatus.name === 'success'
    ) {
      isReady = true
    }

    let statusMessage = ''
    if (video.storageStatus.name !== 'success' && title === null) {
      statusMessage = 'Please provide a title and description'
    } else if (video.transcodingStatus.name === 'success') {
      statusMessage = 'Your video is ready to play'
    } else if (video.transcodingStatus.name === 'failed') {
      statusMessage = 'Your video could not be transcoded'
    }
```

the color and the text are depending on different status

PR #380, opened 2018-03-20
Hiding elements

jrgarou:
- Hiding play thumb button and URL share when the video isn't ready.
- Remove `overflow scroll` from `VideoList`.
- Hiding View/like/dislike from `Play` with a fake conditional(`video.like`)


jrgarou:
How is working
ezgif com-video-to-gif

Now I'm gonna work on this https://github.com/Paratii-Video/paratii-portal/issues/365

Issue #400, opened 2018-03-26
playing a video for the first time is messed up


jellegerbrandy:
@bent0b0x , I have the following issue:

- upload a new video
- publish it
- play it. Now it seems that there are two different vids interacting somehow. You'll notic if you pause the video.

bent0b0x:
I am glad you are able to reproduce this, this sounds just like #252. I'll work on this later today.

PR #409, opened 2018-03-27
400 - always destroy player before creating a new one

bent0b0x:
**Issue**
#400 

**Problem**
The play component could accidentally render two video elements simultaneously, which led to a terrible player experience. The controls would jump around and you would hear two videos while only seeing one!

**Work Done**
Ensure all players are destroyed the moment before creating a new one.

### DIFF
--- a/src/scripts/components/Play.js
+++ b/src/scripts/components/Play.js
@@ -552,9 +552,7 @@ class Play extends Component<Props, State> {
 
   createPlayer = (video: VideoRecord): void => {
     const { updateVolume } = this.props
-    if (this.player && this.player.destroy) {
-      this.player.destroy()
-    }
+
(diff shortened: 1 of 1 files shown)


src/scripts/components/Play.js
```diff
@@ -563,6 +561,10 @@ class Play extends Component<Props, State> {
       poster = video.thumbnails.get(0)
     }
     import('paratii-mediaplayer').then(CreatePlayer => {
+      if (this.player && this.player.destroy) {
+        this.player.destroy()
+      }
+
```
bent0b0x: Since our player creating is aync, we must destroy the player after the dynamic import of `paratii-mediplayer` is complete.

PR #414, opened 2018-03-29
Feat/secure wallet

eliawk:
- Secure keystore flow done and semi tested
- Unlock user keystore w/ pin
- Need UI improvements  
- I created a ton of container, maybe too many duplications

Still missing the restore keystore flow and some refactoring


bent0b0x:
@eliawk took a look. I know this is a WIP but still I think everything here is going in the right direction. Also, there is no problem w/ creating the number of containers you did, I think they are all valid cases. Let me know when this is closer to merging and I can take a more in-depth look!

config/development.json
```diff
@@ -1,6 +1,5 @@
 {
   "provider": "http://localhost:8545",
-  "address": "0x9e2d04eef5b16CFfB4328Ddd027B55736407B275",
-  "privateKey": "399b141d0cc2b863b2f514ffe53edc6afc9416d5899da4d9bd2350074c38f1c6",
```
bent0b0x: Should these be deleted? Just making sure

src/scripts/components/Wallet.js
```diff
@@ -1,13 +1,28 @@
...
+  secureWallet: (e: Object) => void
+
+  constructor (props: Props) {
+    super(props)
+    this.secureWallet = this.secureWallet.bind(this)
+  }
+
+  secureWallet () {
+    console.log('secure')
+    this.props.openModal('ModalSecure')
```
bent0b0x: It's not necessary to do, but if you find yourself with extra time (😜) it would be awesome to not use raw strings and instead move all modal names to a new `constants/ModalConstants.js` file and define all modal names/import them from there.

src/scripts/components/widgets/NumPad.js
```diff
@@ -0,0 +1,120 @@
+import React, { Component } from 'react'
+import styled from 'styled-components'
+import Colors from 'components/foundations/base/Colors'
+
+type Props = {
+  onSetPin: () => String,
```
bent0b0x: Does this actually return a string? Doesn't look like the return value is used in this component.

config/development.json
eliawk: yes, just on dev - because we need to create the wallet from the mnemonic

src/scripts/components/widgets/NumPad.js
eliawk: @bent0b0x i use this pass the pin to the parent component..

bent0b0x: On line 90 though the return value is not used

PR #424, opened 2018-04-03
Player Tests

bent0b0x:
**Problem**
We are severely lacking in functional tests for the player 😬.

**Work Done**
- add some tests for the player

**TODO**
- [x] add a ton more tests
- [x] determine and carry out strategy for being able to execute the tests on circleCI and always have at least one video id that we can depend on loading. Additionally, we must be able to run the tests locally with no issue. @jellegerbrandy @geckoslair, any thoughts?
- [x] write tests for the new functionality in [this PR](https://github.com/Paratii-Video/paratii-portal/pull/453)

**NOTE**
A lot of the changes here are merely adding `data-test-id` attributes to elements


jellegerbrandy:
Yes.  For actually getting the video file, we can either use `nock` (Which we use in paratii.js testing) to mock requests to `gateway.paratii.video` , or, alternatively, instead of the gateway, pass something liek `file:///path/to/video/directory` as the parameter for the gateway to paratii.js. (Let's discuss this in the call)

bent0b0x:
Sounds good. I am not blocked at all, as I can test these new tests against a local video for now, but I will need to figure out how to run these tests on circle before merging.

Could you (or maybe @ya7ya ?) elaborate on how I should use `nock` to mock requests to the gateway? Are there specific requests that I'll need to mock out? If so, is it clear at this point what those requests are and how we can easily mock the responses?

bent0b0x:
Hey @ya7ya or @jellegerbrandy, when you have time could you provide some details/examples as to how we can best mock out the gateway in the portal tests to serve up a local video via our player?

Thanks!

bent0b0x:
If someone could help w/ some advice on how we can get these tests reliably running on circle that would be amazing. 

These tests would have prevented the issue resolved in https://github.com/Paratii-Video/paratii-portal/pull/464

bent0b0x:
This is finally working and ready!

Issue #432, opened 2018-04-05
Rebuild the video title on the embeded player


pedrocasa:
Correct the typography size of the embeddable videos. (Sizes valus are in the Adobe XD dropbox and on the Adobe specs of the player UI-kit.)
Also, we show consider put three small dots if the entire title of the video doesn't fit inside the iframe

jrgarou:
@pedrocasa I can't open those files on Adobe XD.

<img width="350" alt="screen shot 2018-04-05 at 19 24 23" src="https://user-images.githubusercontent.com/320157/38394961-fcc476f8-3906-11e8-8ff5-2eaf6b02a7cf.png">

bent0b0x:
Can we close this one?

jrgarou:
@pedrocasa could you check this?

Issue #421, opened 2018-04-02
Quality control on the player -> cleanup


felipegaucho:
<img width="246" alt="screen shot 2018-04-02 at 2 19 39 pm" src="https://user-images.githubusercontent.com/26517120/38206484-ee551c5e-3680-11e8-8012-90db204265c3.png">

- [x] Video quality -> Quality
- [x] kbps, or px -> p (simply p, as in `1080p`)
- [x] Automatic -> Auto.
- [x] Scrollbar on Mac -> @jrgarou plz help us kill it, @bent0b0x can't test it since he has no Mac, we can even hide some resolutions in order to avoid the pop up from becoming too long vertically.

jrgarou:
- [ ] Remove scrollbar on `/embed

![quality-modal-generates-a-scroll](https://user-images.githubusercontent.com/320157/38211779-f6b5371a-3691-11e8-96a7-bd2f04d8bab7.gif)

felipegaucho:
@jrgarou, in case it's still not clear: we do not want the scroll bar to be there. It is a bug. We need you to help take it away. We should never have more resolutions than what fits inside the box.

🇧🇷 Nunca devemos ter mais resoluções do que cabe na caixa - não queremos a scroll bar jamais, nunca, em nenhum lugar dentro do player.

jrgarou:
@felipegaucho it's clear 👍
My comment is just another fix for the "Quality control".

ya7ya:
hey , regarding the video qualities , here are the `MAX_BIT_RATES` i use in our transcoder. note that this isn't an exact science, and these values could change.

```javascript
const MAX_BIT_RATE = {
  '1080': '2760700',
  '720': '1327100',
  '480': '763900',
  '360': '423900',
  '240': '155600',
  '144': '64000'
}
```

bent0b0x:
Can we close this @felipegaucho @jrgarou ?

jrgarou:
it's ok to me

felipegaucho:
Superb @jrgarou 👏

PR #462, opened 2018-04-10
Set a limit to "Video Quality" height 

jrgarou:
- Hide the last elements if out of bounds #421 
https://github.com/Paratii-Video/paratii-portal/blob/9a12e7806a4cd448f4c585cda2940cc2e7a44cf7/src/scripts/components/widgets/PlaybackLevels.js#L127-L139
https://github.com/Paratii-Video/paratii-portal/blob/9a12e7806a4cd448f4c585cda2940cc2e7a44cf7/src/scripts/components/widgets/PlaybackLevels.js#L157
image

- Player title #432 
https://github.com/Paratii-Video/paratii-portal/blob/9a12e7806a4cd448f4c585cda2940cc2e7a44cf7/src/scripts/components/VideoOverlay.js#L109-L111


bent0b0x:
@jrgarou please delete `package-lock.json`. We are no longer using it; only need `yarn.lock`

PR #464, opened 2018-04-12
Do not overwrite video id

bent0b0x:
**Problem**
No videos are playing on staging/local. This is because the FE is expecting the video ID it gets from `paratii-js` to be present at key `_id`. However, due to what I think is a recent change in `paratii-js`, we can expect to find the id simply at key `id`. 

**Work Done**
Drop usage of `_id` in favor if `id`, and now videos play.


jellegerbrandy:
yeah, this indeed had just changed

Issue #475, opened 2018-04-14
Remove hard-coded portal urls


bent0b0x:
**Problem**
We have hard-coded `portal.paratii.video` in a few places. This prevents us from, under certain circumstances (e.g. on telegram and probably many more) testing our embeds on staging. 

Essentially we embed a video that exists on staging but ultimately the production embed bundle is loaded, and these two will not be compatible (two problems: we load the wrong code and secondly the id from a staging video cannot be found on production).

We should remove these hard-coded urls so that we can test embeds on staging and locally more easily.

PR #476, opened 2018-04-14
475 - remove hard-coded urls

bent0b0x:
**Issue**
#475 

**Problem**
We have hard-coded urls in our app that make testing embeds (and other things) on staging difficult or impossible.

**Work Done**
- remove all instances of hard-coded `portal.paratii.video` urls from our app
- add a util that gets the right url based on the environment

### DIFF
--- a/src/scripts/components/Play.js
+++ b/src/scripts/components/Play.js
@@ -18,7 +18,11 @@ import Text from 'components/foundations/Text'
 import Card from 'components/structures/Card'
 import ShareOverlay from 'containers/widgets/ShareOverlayContainer'
 import VideoNotFound from './pages/VideoNotFound'
-import { requestFullscreen, requestCancelFullscreen } from 'utils/AppUtils'
+import {
+  requestFullscreen,
+  requestCancelFullscreen,
+  getAppRootUrl
+} from 'utils/AppUtils'
 import { PLAYER_PARAMS } from 'constants/PlayerConstants'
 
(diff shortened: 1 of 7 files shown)


src/scripts/components/Play.js
```diff
@@ -629,49 +633,54 @@ class Play extends Component<Props, State> {
     return !valueIsFalse
   }
 
-  getPortalUrl () {
-    // FIXME: do not hardcode this here
```
bent0b0x: Nice to fix `FIXME` comments :)

src/scripts/utils/AppUtils.js
```diff
@@ -15,6 +15,20 @@ export const getRoot = (): Element => {
   return root
 }
 
+export const getAppRootUrl = (env: ?string = 'development'): string => {
+  switch (env) {
+    case 'production':
+      return 'https://portal.paratii.video'
+    case 'test':
+      return 'https://staging.paratii.video'
```
bent0b0x: Maybe this should also be `localhost`...but not sure that it matters

src/scripts/utils/AppUtils.js
```diff
@@ -15,6 +15,20 @@ export const getRoot = (): Element => {
   return root
 }
 
+export const getAppRootUrl = (env: ?string = 'development'): string => {
+  switch (env) {
+    case 'production':
+      return 'https://portal.paratii.video'
```
bent0b0x: Should probably move these to constants

Issue #449, opened 2018-04-09
Player start screen/state


jrgarou:
We need a start screen or state for the player when the `autoplay` is `false/0`
State:  We could just show the play controls and make it hide only when the video starts to play(fast, I think).
Screen: A design that calls the user to click on the screen to start the video. Perhaps a play button in the middle of the screen(video canvas).

PR #482, opened 2018-04-19
Start Screen #449

jrgarou:
Player start screen/state #449 

player-states


src/scripts/components/Play.js
```diff
@@ -584,6 +588,12 @@ class Play extends Component<Props, State> {
       } else {
         player.play()
       }
+
+      if (this.state.isStartScreen) {
+        this.setState({
+          isStartScreen: false
```
bent0b0x: Do we show the start screen basically anytime the video is not playing? If that is the case, then we should just remove `isStartScreen` from state and base whether we show the start screen solely off of `this.props.isPlaying`. This will avoid data duplication.

Do you see something wrong w/ this approach? Even if we can't do it _entirely_ from `this.props.isPlaying` I'd bet we can derive whether to show the start screen from that prop and maybe another.

src/scripts/components/VideoOverlay.js
```diff
@@ -40,67 +41,69 @@ type Props = {
...
 }
 
+const Z_INDEX_OVERLAY: string = '1'
+const Z_INDEX_CONTROLS: string = '2'
+const Z_INDEX_OVERLAY_SHADOW: string = '1'
+const Z_INDEX_CENTRALIZEDCONTENT: string = '2'
+const Z_INDEX_TITLE: string = '3'
+const Z_INDEX_BUTTONS: string = '4'
+
+const UpDownElements = css`
```
bent0b0x: What is `UpDownElements`? Wondering if there is a more descriptive name we might be able to use

src/scripts/components/Play.js
jrgarou: > Do we show the start screen basically anytime the video is not playing?

No. Just in embed. One time only.
I don't understand this approach

bent0b0x: My point is, why do we have `this.state.isStartScreen`? Can't the logic as to whether we show the screen just be `if (this.props.isEmbed && this.props.isPlaying)`?

bent0b0x: You could write a method like this:

```javascript
shouldShowStartScreen() {
  return this.props.isEmbed && this.props.isPlaying
}
```

Assuming that is indeed the correct logic

jrgarou: The `this.props.isEmbed && this.props.isPlaying` could be a *paused video state* or a *ended video state*. And this *start state* is different from both.

bent0b0x: @jrgarou I think my concerns here were legitimate. If you visit the `/embed` url with `autoplay=true`, you'll see that the start screen does not go away. I will work on a fix now and tag you in PR

Issue #503, opened 2018-04-26
Search Video - secure account


eliawk:
- You are logged in
- You search a video ("test")
- You click a video
- The app reload the url -> re ask the password

PR #505, opened 2018-04-26
130 - remove anchor tag in favor of react router link

bent0b0x:
**Issue**
#503 

**Problem**
When clicking on a search result, we reload the entire app. Instead, we want to be more SPA-like and just load the correct route w/o a page reload. This helps many things in the UX, including the secure account flow.

cc @eliawk


eliawk:
Yes, thanks @bent0b0x !

Issue #130, opened 2018-02-09
Create search / discovery page


felipegaucho:
I'd suggest 'modules' from which one (or a few) can be placed in the current portal/uploader homepage, in landing pages, or as a window (a single module) on third-party sites.

- [ ] Design it
- [ ] Borrow existing search logic from the lib
- [ ] Implement it

jellegerbrandy:
This is directly related to this issue in paratii-db: https://github.com/Paratii-Video/paratii-db/issues/26

pedrocasa:
I have sent the first sketches of the search page for @geckoslair (images below)
_there are some small differences on each picture_

![upload-1920-_-9](https://user-images.githubusercontent.com/30123644/38741664-c83784c4-3f10-11e8-8a87-c885c591c776.png)
![upload-1920-_-11](https://user-images.githubusercontent.com/30123644/38741665-c85875ee-3f10-11e8-8ad6-dc98c329ba17.png)


After he saw it he sent me a list of suggestions and notes about, I really like some of them and I'm already designing it with those changes for u

bent0b0x:
From an engineering perspective, I think we need to keep a few things in mind when we get started on this.

@pedrocasa @felipegaucho I’m assuming we will eventually want filters on this page, yes? At least free vs paid.

We should make sure the we build out this feature in the UI in a way so that it is not extremely painful to account for those future enhancements. 

In addition, we should ensure we have some sort of pagination mechanism put into place so that we can fetch some videos and then on scroll down, for example, fetch the next set of results until.

@jellegerbrandy do you know if the search in the lib currently supports paginated results? I will check the shiny new docs now and see if I can find the answer :)

jellegerbrandy:
We have paginated results now (since a few days, they are used on staging
but not in production yet). The docs are not updated yet,
https://github.com/Paratii-Video/paratii-js/issues/204

geckoslair:
@felipegaucho @jellegerbrandy @bent0b0x @pedrocasa for me it's ok in a paratii-db perspective =)
i suggest just to have better crunched images for thumbnails since actuals are tooooo heavy.

pedrocasa:
More evolved version of it https://xd.adobe.com/spec/6e9bbad0-7d91-41e4-bc51-a1f31ff57882/#
Still working on an alt version using the square grid system ( so we can be sure about the decision of building it using rows)

bent0b0x:
[...]

PR #506, opened 2018-04-26
130 - add loader for when search results are being loaded

bent0b0x:
**Issue**
#130 

**Work Done**
- create generic loading component
- when search results are loading show a spinner

**Demo**

searchloader


src/server/views/index.hbs
```diff
@@ -1,38 +1,31 @@
...
-      {{else}}
-        {{> notfoundmetas }}
-      {{/if}}
-    {{/if}}
 
+<head>
+  <meta charset="utf-8">
+  <title>{{title}}</title>
+  <meta name="viewport" content="width=device-width,user-scalable=no"> {{#if player}} {{#if video}} {{> metas }} {{else}} {{> notfoundmetas }} {{/if}} {{/if}}
+
```
bent0b0x: Just prettier changes

src/server/views/index.hbs
```diff
@@ -1,38 +1,31 @@
...
+  <style>
+  </style>
+</head>
+
+<body>
+  <noscript>
+    You need to enable JavaScript to run this app.
+  </noscript>
+  <div id="root">
+    <span class="app-loader paratii-loader" style="width: 50px; height: 50px;"></span>
```
bent0b0x: Moved this width/height from the stylesheet(s) to the html so that we can have better control over the size of our loaders

PR #507, opened 2018-04-26
449 - don't show start screen if we are playing for any reason

bent0b0x:
**Issue**
#449 

**Problem**
When a video played because of autoplay, the start screen would not go away

autoplayfail

**Work Done**
Hide the start screen if the video plays for any reason (including autoplay)

autoplaywin


bent0b0x:
@jrgarou feel free to give this a test before merging since you are most familiar w/ the expected behavior

Issue #501, opened 2018-04-26
PLAYER - PARATII -- "play" icon and timeline bar


jellegerbrandy:
Toddy tells us: https://github.com/Paratii-Video/paratii-js/issues/222

> hey guys, icon "play" just disappear sometimes in the beginning of timeline bar , we should decrease the size of timeline bar, may this issue will be fix.

jellegerbrandy:
@jrgarou @bent0b0x @pedrocasa

jrgarou:
This happens because of this buttons not use our `MainSVG`. 
It loads an SVG file with `mask-image`.
I think that all icons from interface should be a `<svg><use></svg>` [tag](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use).

(the notifications icons have the same issue)

jrgarou:
@bent0b0x 
I'll update some SVG icons with [SVGIcon](https://github.com/Paratii-Video/paratii-portal/blob/dev/src/scripts/components/foundations/Button.js#L84) and I make the same with those icons.

felipegaucho:
Should we close this @jellegerbrandy @jrgarou ?

Issue #499, opened 2018-04-25
it is unclear how to copy the 12 words


jellegerbrandy:
I see  that you can click to copy, but it seems to me that we do not need to invent new ways of copy-pasting here. They should be selectable (as well).

pedrocasa:
There are two ways to solve it

Create a copy button as we did in the past
-
![screen shot 2018-04-26 at 12 05 38 pm](https://user-images.githubusercontent.com/30123644/39314230-2a5d523a-494a-11e8-9804-31ad6b879ecb.png)

Good points of it 
- Still using the number boxes can show the 12 words in a very explicit way, it clear what the order is and it's easier to copy find each word if you are writing it down in a piece of paper

Bad points of it
- It brings one more element to the layout. We already have the word boxes, the checkbox and now and now we are adding a button to copy it


Not using the word boxes 
-

![image](https://user-images.githubusercontent.com/30123644/39313702-0580314a-4949-11e8-9c73-d0340bbcd615.png)

Good points of it
- Easier to copy and paste because it's a text line and the user will probably think that works in the same way as every text line on the internet ( he would think it's copiable)

jellegerbrandy:
In spanckchain (I think) you the words are shown in boxes, but you can just select them with your mouse. I.e. it behaves like normal text.

pedrocasa:
well,
I was mistaken about the good point of the first picture. I've made some real tests here at Bossa Nova to see how people would copy it on a piece of paper and the result was not what I was expecting. 

![screen shot 2018-04-26 at 3 14 29 pm](https://user-images.githubusercontent.com/30123644/39324020-c3eaa9f6-4964-11e8-80a4-f3b67765b6c9.png)

- People took twice the time to copy the 12 words in the first scenario.
-  There are some people (including one that already knows a lot about crypto) that copied the 12 words like this ( image below). This is terrible because they may try to enter on their account using a completely wrong word order

![image](https://user-images.githubusercontent.com/30123644/39323462-12740ea2-4963-11e8-9b31-8ae6451d665a.png)

I do think that we should use a layout like this (image below). I'm not a big fan of how it looks in comparison to the word-box layout, but it's the best choice in the usability matter

jrgarou:
Nice, @pedrocasa 

A copy button would be quite user-friendly.

![image](https://user-images.githubusercontent.com/320157/39326354-15ddb2a2-496b-11e8-806c-33b3f3502fb0.png)

[...]

PR #511, opened 2018-04-27
Showseed flow + svg icons - #501 #499

jrgarou:
- Change `ModalShowSeed` UI https://github.com/Paratii-Video/paratii-portal/issues/499#issuecomment-384768291
- Use ` <SVGIcon />` instead of `<use/>` tag (include https://github.com/Paratii-Video/paratii-portal/issues/501#issuecomment-384550101)
https://github.com/Paratii-Video/paratii-portal/blob/f5ac48a97fc37680a2b76f122b274d35d9ca5a25/src/scripts/components/foundations/Button.js#L88-L99
- Changes SVGs from Secure Flow


src/scripts/components/Play.js
```diff
@@ -12,7 +12,7 @@ import queryString from 'query-string'
 import { PlaybackLevel } from 'records/PlayerRecords'
 import VideoRecord from 'records/VideoRecords'
 import VideoOverlayContainer from 'containers/VideoOverlayContainer'
-import Button from 'components/foundations/Button'
+import Button, { SVGIcon } from 'components/foundations/Button'
```
bent0b0x: Why is `SVGIcon` part of the `Button.js` file? I think we should separate it. I wouldn't think svg icons need to be coupled to buttons

src/scripts/components/Wallet.js
```diff
@@ -10,9 +9,9 @@ import { ACTIVATE_SECURE_WALLET } from 'constants/ParatiiLibConstants'
 type Props = {
   user: Object,
   userAddress: string,
-  isWalletSecured: boolean,
+  isWalletSecured: string,
```
bent0b0x: Hm why is this a `string`? If we are passing in a `string` I think _that_ is the problem. We shouldn't change this flow type

jrgarou: this is from @eliawk 
In fact wasn't to be here. Let me change

src/scripts/components/Play.js
jrgarou: Totally agree, @bent0b0x

src/scripts/components/Wallet.js
```diff
@@ -12,7 +11,7 @@ type Props = {
   userAddress: string,
   isWalletSecured: boolean,
   closeModal: () => void,
-  openModal: string => void
+  openModal: () => void
```
eliawk: openModal need the string of the modal you want to open. why do you change w/ : `() => void`

PR #513, opened 2018-04-27
Allow ask for password to be submitted by "Enter" key

bent0b0x:
**Work Done**
- Turn ask for password modal into a form so that you can submit it by pressing "Enter" or clicking the "continue" button

**TODO**
- [ ] update the other modals in the secure account flow


src/scripts/components/widgets/modals/ModalAskPassword.js
```diff
@@ -101,7 +104,6 @@ class ModalAskPassword extends Component<Props, Object> {
         },
         'error'
       )
-      throw err
```
bent0b0x: @eliawk I don't _think_ we want to throw errors here; this is polluting the browser console and we are, in fact, handling these errors. What are your thoughts?

src/scripts/components/widgets/modals/ModalAskPassword.js
```diff
@@ -144,8 +146,9 @@ class ModalAskPassword extends Component<Props, Object> {
             <ButtonContainer>
               <Button
                 data-test-id="continue"
+                form={FORM_ID}
+                type="submit"
                 purple
-                onClick={this.setPassword}
```
bent0b0x: Now this is just handled via the `onSubmit` handler by the form

src/scripts/components/widgets/modals/ModalAskPassword.js
eliawk: yes sounds good to me

Issue #515, opened 2018-04-27
Password feedback is easy to miss and can be irritating 


bent0b0x:
I saw someone else mention this, but although we tell the user upfront about the qualification for the password it is easy to miss and it would be better to get clear feedback on the user's password  before they try to submit it.

![passwordfail](https://user-images.githubusercontent.com/7697924/39387659-d3235a28-4a48-11e8-8899-1baee7c57cb6.gif)

felipegaucho:
Closing it, tks @bent0b0x https://github.com/Paratii-Video/paratii-portal/pull/516

PR #516, opened 2018-04-28
515 - Password validation UX

bent0b0x:
**Issue**
#515 

**Work Done**
Now, password validation works as follows:
- we show multiple validation errors at once, if there are more than one
- we show errors with password validation as soon as the user moves on to the confirm input (so they do not have to wait until submitting to see errors). After they have moved to the confirm input, the errors will remain until they are fixed
- we should "passwords do not match" errors after the user tries to submit their password, and then continues to show it until they match
- allow submitting passwords by pressing "enter"
- there is extra margin between the inputs and the buttons to leave room for errors, so that when errors appear the modal does not quickly grow in size (@pedrocasa open to feedback on this change)

**Demo**

passwordwin

@pedrocasa @eliawk (and anyone else) let me know what you think of these proposed changes!


jellegerbrandy:
@eliawk , @bent0b0x , I'm not sure about how to resolve this conflict, can you help me?

```
<<<<<<< HEAD
import { getPasswordValidationErrors } from 'utils/AppUtils'
import BigLockSvg from 'components/foundations/svgs/BigLockSvg'
=======
import { passwordStrength } from 'utils/AppUtils'
import PadlockSvg from 'components/foundations/svgs/PadlockSvg'
>>>>>>> dev
```

PR #521, opened 2018-04-29
218 - Fix paginated searching

bent0b0x:
⚠️ DEPENDENT ON https://github.com/Paratii-Video/paratii-js/pull/226 ⚠️ 

**Issue**
#218 

**Work Done***
- update types of params passed to `.search()` call
- fix logic for paginating search results in the search reducer


jellegerbrandy:
I've merged this perhaps too quickly. If this  depends on a change in paratii-js,  should we not have an updated paratii-js dependency in this PR? Or, otherwise, a failing test?

bent0b0x:
Yeah I believe we need to release a new `paratii-js` and then update the portal dependency

PR #529, opened 2018-05-02
Login - Signup

eliawk:


src/scripts/actions/UserActions.js
```diff
@@ -117,9 +117,9 @@ export const setupKeystore = () => async (
 
   // Case 2: we have a secured wallet is localStorage
   if (walletStringSecure && ACTIVATE_SECURE_WALLET) {
-    console.log('Try to open encrypted keystore')
+    // console.log('Try to open encrypted keystore')
     // Need to ask the PIN
-    dispatch(openModal(MODAL.ASK_PASSWORD))
+    // dispatch(openModal(MODAL.ASK_PASSWORD))
```
bent0b0x: remove commented lines?

test/functional-tests/app.spec.js
```diff
@@ -7,12 +7,12 @@ describe('Portal application', () => {
...
-    })
-  })
+  // describe('Nav bar', () => {
+  //   it("should render the user's PTI balance", () => {
+  //     browser.url('http://localhost:8080')
+  //     browser.waitUntil(() =>
+  //       browser.getText('[data-test-id="nav-pti-balance"]').includes('PTI')
+  //     )
+  //   })
+  // })
```
bent0b0x: Do we need to fix these tests?

test/functional-tests/profile.spec.js
```diff
@@ -459,94 +418,94 @@ describe('Profile and accounts workflow:', function () {
...
+  //     browser.click('#logoutBtn')
+  //     browser.waitForClickable('#nav-profile')
+  //     assertUserIsNotLoggedIn(browser)
+  //
+  //     login(browser, 'foobar')
+  //
+  //     waitForUserIsLoggedIn(browser)
+  //     assertUserIsLoggedIn(browser)
+  //   })
+  // })
```
bent0b0x: Are these just temporarily commented out?

eliawk: we are not using so maybe we could remove it but it's a temporary reminder

test/functional-tests/app.spec.js
eliawk: now when you arrive to the app there is no more the `PTI balance`, so this test don't make sense

src/scripts/actions/UserActions.js
eliawk: it's temporary, because now i've removed the `askPassword` at the beginning but maybe it will come back|

bent0b0x: I would just delete it, we can easily add it back in later

test/functional-tests/app.spec.js
bent0b0x: Can we delete them then?

test/functional-tests/profile.spec.js
bent0b0x: If we are _sure_ we are going to need this later I'd just add a `.skip`, otherwise delete

Issue #544, opened 2018-05-06
Autoplay in portal not always working


bent0b0x:
- visit https://staging.paratii.video/play/TXWzBRXgWytP
- the video does not play w/o user interaction (maybe it will sometimes, there could be a race-case)

bent0b0x:
This is almost surely due to new restrictions in chrome when it comes to autoplay (https://developers.google.com/web/updates/2017/09/autoplay-policy-changes).

Safari has also implemented similar restrictions, in that by default (I believe) autoplay is disabled on all pages.

We need to detect if autoplay fails and then show the start screen if so. So far the clappr event system doesn't seem to help with this, but that is what our solution should be (most likely).

bent0b0x:
Currently, the issue is that when we try to play the video w/ clappr and chrome and/or safari blocks it, I'm not sure if we get notified by an event that the player has stopped. Thus, we can't respond and show the start screen.

One quick patch is to just not autplay videos ever for now

bent0b0x:
This seems broken again

jellegerbrandy:
sems a browser issue

PR #549, opened 2018-05-06
544 - autoplay UX fix

bent0b0x:
**Issue**
#544 

**Work Done**
- allow showing of start scree (big play button) in portal player
- when autoplay fails (for whatever reason, but most likely because safari/chrome have blocked it for various reason), keep the start screen visible so that user knows to manually play video

**Demo**

autoplaywin

☝️ previously, the start screen/play button would disappear after autoplay is blocked

cc @pedrocasa


jrgarou:
Nice.
I think the _play button_ should be visible/enable when there's a video to play. If the _play button_ doesn't trigger any action we should hide it until so.

bent0b0x:
@jrgarou do you mean the big play button or the play button on the controls bar?

Can you elaborate a bit more on a particular situation when you think we should hide it when we are not currently hiding it?

jrgarou:
@bent0b0x if there's no situation when the big play button and controls are visible and the video is not ready I think is fine.

jellegerbrandy:
can you fix the conflict? thks

bent0b0x:
On it

bent0b0x:
@jrgarou I _think_ we should be fine here. If the video hasn't loaded yet clicking play should trigger clappr to play once it has buffered enough. So the play button should trigger an action always even if we aren't ready to play immediately.

I will add a condition to not show the button until we have loaded a clappr player, I think that will cover our bases

jellegerbrandy:
did this now go in conflict with your other PR I just merged? :-)

jrgarou:
@bent0b0x we can define this here:
https://github.com/Paratii-Video/paratii-portal/blob/ec88c867bb57ebae8c5420cc738f718dd698dcaf/src/scripts/components/Play.js#L749-L753

bent0b0x:
@jellegerbrandy fixed!

Issue #548, opened 2018-05-06
Play page layout is broken in firefox


bent0b0x:
Probably a flexbox height issue

![image](https://user-images.githubusercontent.com/7697924/39674430-c62429d8-5119-11e8-8a99-06618fa258b7.png)

PR #550, opened 2018-05-06
548 - firefox layout fix

bent0b0x:
**Issue**
#548 

**Work Done**
- fix aspect ratio css to work in firefox

**Before**

image

**After**

firefoxwin

### DIFF
--- a/src/scripts/components/Play.js
+++ b/src/scripts/components/Play.js
@@ -87,10 +87,18 @@ const Wrapper = styled.div`
 `
 
 const VideoWrapper = styled.div`
-  margin: ${props => (props.isEmbed ? null : '0 auto 25px')};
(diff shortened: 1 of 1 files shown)


src/scripts/components/Play.js
```diff
@@ -72,6 +72,7 @@ const Wrapper = styled.div`
   flex-direction: column;
```
bent0b0x: Basically all this is is adding a new `div` around the video element; apparently the `padding-bottom` aspect ratio css trick doesn't work in firefox unless you do this

jellegerbrandy:
@bent0b0x this went into conflict with another PR, can you check?

bent0b0x:
Fixed merged conflict

Issue #547, opened 2018-05-06
SVG continuously requested as video plays


bent0b0x:
When a video plays, the browser requests `card-bg.svg` multiple times a second. This cannot be good for performance.

We should handle his SVG as we do other SVGs so this no longer happens. It should only be loaded 1x.

**Demo**

![svgfail](https://user-images.githubusercontent.com/7697924/39674401-6516013e-5119-11e8-9b59-8fdf81952db8.gif)

cc @jrgarou

jrgarou:
I agree. But why these requests are happening, @bent0b0x?

jrgarou:
Anyway, this `Card` should have the `nobackground` prop

bent0b0x:
Yeah, I'm not really sure why were were making all of these requests but for now the attached PR is a quick fix

PR #551, opened 2018-05-06
547 - load svg inline so that it is not requested for multiple times

bent0b0x:
**Issue**
#547 

**Work Done**
- load svg data url instead of having browser potentially request asset many many times

### DIFF
--- a/src/scripts/components/structures/Card.js
+++ b/src/scripts/components/structures/Card.js
@@ -2,6 +2,8 @@ import React, { Component } from 'react'
 import BaseTitle from '../foundations/Title'
 import styled from 'styled-components'
 
+import cardBackgroundSvgDataUrl from 'assets/svg/card-bg.svg'
+
 type Props = {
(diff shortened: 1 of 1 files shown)


src/scripts/components/structures/Card.js
```diff
@@ -64,7 +66,7 @@ export const CardWrapper = styled.div`
 
 const Main = styled.div`
   background: ${props => props.theme.colors.MainCard.background}
-    ${props => (props.nobackground ? null : "url('assets/svg/card-bg.svg')")}
+    ${props => (props.nobackground ? null : `url(${cardBackgroundSvgDataUrl})`)}
```
bent0b0x: @jrgarou can you help explain why we need this svg background at all? It seems to just be a solid color background anyways

jrgarou: After a lot of changes, there're few cards with the background - or no one maybe.

PR #565, opened 2018-05-09
Edit Profile

jrgarou:
Edit profile page
(@eliawk @jellegerbrandy  The `paratii.users.get/update` is still returning `null`)


eliawk:
@jrgarou the PR looks good, but the `paratii.users.update` is not working (need to fix this in `paratii.js`).

the error is this 
```
Uncaught (in promise) TypeError: Cannot set property 'name' of null
    at ParatiiUsers.update (paratii.core.users.js?abbd:84)
```

instead `paratii.users.get()` is working good. it returns null id the user don't have any informations (username, email)

jellegerbrandy:
can you make an issue on paratii-js that is a bit more precise than "is not
working"?

Issue #571, opened 2018-05-11
Make username required in account secure flow


bent0b0x:
Right now, you can dismiss this modal (by clicking X or by clicking outside of the modal)

![image](https://user-images.githubusercontent.com/7697924/39934322-720ecb66-5513-11e8-93b7-bd3713ec2564.png)

Let's not make this optional, because we want reasonable usernames for video authors/creators throughout the system.

pedrocasa:
On my first designs, it wasn't optional.
If someone could fix it would be nice

felipegaucho:
Tks @bent0b0x 👊

PR #581, opened 2018-05-16
571 - require username in account flow

bent0b0x:
**Issue**
#571 

**Work Done**
- prevent user from closing username/email modal until they have entered a username

**Demo**

usernamewin

### DIFF
--- a/src/scripts/components/widgets/modals/Modal.js
+++ b/src/scripts/components/widgets/modals/Modal.js
@@ -112,27 +112,13 @@ export const ModalScrollContent = styled.div`
 `
 
 class Modal extends Component<Props, void> {
-  renderModal () {
-    const { modalContent } = this.props
-    switch (modalContent) {
-      case MODAL.STAKE:
-        return <ModalStake />
-      case MODAL.SECURE:
-        return <ModalSecure />
-      case MODAL.GREAT_POWER:
-        return <ModalGreatPower />
-      case MODAL.CREATE_PASSWORD:
-        return <ModalCreatePassword />
-      case MODAL.ASK_PASSWORD:
-        return <ModalAskPassword />
-      case MODAL.SHOW_SEED:
-        return <ModalShowSeed />
-      case MODAL.PROFILE:
-        return <ModalProfile />
-      case MODAL.REWRITE_SEED:
-        return <ModalRewriteSeed />
-      case MODAL.RESTORE_ACCOUNT:
-        return <ModalRestoreAccount />
+  canClose = () => {
+    return this.props.modalContent !== MODAL.PROFILE
+  }
+
+  close = () => {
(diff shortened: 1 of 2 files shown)


src/scripts/components/widgets/modals/Modal.js
```diff
@@ -112,27 +112,13 @@ export const ModalScrollContent = styled.div`
 `
 
 class Modal extends Component<Props, void> {
-  renderModal () {
```
bent0b0x: I just moved this method

src/scripts/components/widgets/modals/Modal.js
```diff
@@ -112,27 +112,13 @@ export const ModalScrollContent = styled.div`
...
-      case MODAL.SHOW_SEED:
-        return <ModalShowSeed />
-      case MODAL.PROFILE:
-        return <ModalProfile />
-      case MODAL.REWRITE_SEED:
-        return <ModalRewriteSeed />
-      case MODAL.RESTORE_ACCOUNT:
-        return <ModalRestoreAccount />
+  canClose = () => {
+    return this.props.modalContent !== MODAL.PROFILE
```
bent0b0x: You can no longer close the profile modal. The only way to dismiss it is to enter a username (and email if you want) and then continue to the next step.

src/scripts/components/widgets/modals/ModalProfile.js
```diff
@@ -93,24 +90,26 @@ class ModalSetPassword extends Component<Props, Object> {
...
+            <TextField
+              error={this.state.error.length > 0}
+              label="Email"
+              name="email"
+              type="text"
+              value={this.state.email}
+              onChange={e => this.handleInputChange('email', e)}
+              margin="0 0 30px"
+            />
+          </form>
```
bent0b0x: Wrapped this in a `form` so that you can submit this modal by pressing `enter`

src/scripts/components/widgets/modals/Modal.js
bent0b0x: Yes that is still possible. I think to solve that we would have to rethink the flow a bit and reorder/combine the steps.

For now I think we want to leave the flow as is, plus most users would not do that anyway. But tagging @felipegaucho and @pedrocasa just to get their thoughts.

eliawk: ok @bent0b0x maybe merge this PR and the we refine the flow

jellegerbrandy: agree. I think that data integrity like this should not be an argument to change the flow here. It is always possible that the user stops halfway (e.g. his battery fails); what we should do is think of policies to recover from that. For example, for this case, we can use a default user name, or, on the next visit, ask the user to provide a username before doing anything else.

PR #592, opened 2018-05-22
New upload video

eliawk:
schermata 2018-05-22 alle 15 17 42


eliawk:
at the moment the `Cancel` button has been removed because we need to implement the Cancel Upload on `paratii.js`

src/scripts/components/App.js
```diff
@@ -5,8 +5,6 @@ import { Route, Switch, Redirect } from 'react-router-dom'
 import DocumentTitle from 'react-document-title'
 import { ThemeProvider } from 'styled-components'
 
-import SignupContainer from 'containers/SignupContainer'
-import LoginContainer from 'containers/LoginContainer'
```
bent0b0x: Nice 🙏

src/scripts/components/UploadListItem.js
```diff
@@ -0,0 +1,567 @@
...
+  handleHeight (e: Object) {
+    let height = '0px'
+    if (this.formWrapperRef && this.state.open) {
+      height = this.formWrapperRef.offsetHeight + 'px'
+    }
+    return height
+  }
+
+  render () {
+    const { video, videoId } = this.props
```
bent0b0x: This file and render method are huge. Could it be broken up?

`getStatusMessage` should be something like an `operator` or `util` and live outside of this component. Other parts of `render` can go into helper `renderXXX` methods. Will be easier to reason about this component later if this refactor is doen

src/scripts/components/UploadListItem.js
```diff
@@ -0,0 +1,567 @@
...
+              <VideoMediaIcon>
+                <SVGIcon color="white" icon="icon-player-play" />
+              </VideoMediaIcon>
+              {durationBox}
+            </VideoMediaOverlay>
+            <VideoImage source={videoPoster} />
+          </VideoMediaLink>
+        </VideoMedia>
+      )
+    } else if (isPublishable) {
```
bent0b0x: Generally a good practice to not add this much logic into `render() {}`. Also, instead of using `let` variables in `render` it is more conventional (and easier to read) if you do this:

```javascript
<div>
  { isPublished ? <something /> : <somethingElse /> }
</div>
```

☝️ but the above should still go in a separate render method or even component

src/scripts/components/VideoForm.js
```diff
@@ -1,5 +1,3 @@
-/* @flow */
```
bent0b0x: Why removed `flow`?

src/scripts/components/widgets/FilesUploader.js
```diff
@@ -142,43 +148,65 @@ class FilesUploader extends Component<Props, Object> {
...
+            <UploadCoverText>
+              <UploadCoverTextBig>Drag & drop to upload</UploadCoverTextBig> or
+              choose a file
+            </UploadCoverText>
+          </UploadCover>
+        </UploaderWrapper>
+      )
+    }
+
+    return uploaderBox
```
bent0b0x: Can we avoid `let` in `render`?

Maybe favor `{ showCard ? <something /> : <else />}`

src/scripts/components/UploadListItem.js
eliawk: Agree, but we were in a hurry to finish this

src/scripts/components/VideoForm.js
eliawk: because we are not using this component anymore, but i've given flow error. It's just there until we merge this, because maybe there is some useful code inside

Issue #609, opened 2018-05-23
show flag "verified" in UX


jellegerbrandy:
As specified in #510

jellegerbrandy:
the  flag is on the last page of the design referenced here: https://github.com/Paratii-Video/paratii-portal/issues/590
It does depend on the db returning the flagged value, something that @geckoslair is working on now

PR #625, opened 2018-05-26
609 - add email verify notification and checkmark

bent0b0x:
⚠️ I was not able to actually get this flow to work; the gif below is using stubbed in data ⚠️ 

**NOTES**
- is it expected that when I verify my email I have to restore my account via a mnemonic? I would expect that I could just enter my password.

**Issue**
#609 

**Work Done**
- move email in profile page
- add checkmark next to verified emails
- add notification when you hit `/verify` route, on success or error

image

verifywin

### DIFF
--- a/src/scripts/operators/UserOperators.js
+++ b/src/scripts/operators/UserOperators.js
@@ -1,0 +1,10 @@
+/* @flow */
+
+import User from 'records/UserRecords'
+
+export const getEmail = (user: User): string => user.get('email')
+
+export const getEmailIsVerified = (user: User): boolean =>
+  user.get('emailIsVerified')
+
+export const getName = (user: User): string => user.get('name')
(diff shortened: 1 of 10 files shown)


src/scripts/operators/UserOperators.js
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+
+import User from 'records/UserRecords'
+
+export const getEmail = (user: User): string => user.get('email')
+
+export const getEmailIsVerified = (user: User): boolean =>
+  user.get('emailIsVerified')
+
+export const getName = (user: User): string => user.get('name')
```
bent0b0x: I discussed this w/ Elia, but I like the idea of using `operators` in our codebase. Essentially they take a record as an input and return any output. Useful in selectors and components alike.

src/scripts/records/UserRecords.js
```diff
@@ -25,6 +26,7 @@ class User extends Immutable.Record({
   address: string
   name: string
   email: string
+  emailIsVerified: boolean
```
bent0b0x: @geckoslair can you confirm that this is the correct field name?

Issue #589, opened 2018-05-21
Video Time passes the total duration of the video 0.o


lupape:
The size flagged in the video is 0:44 and the video goes to 0:45.
<img width="827" alt="screen shot 2018-05-21 at 12 23 23 pm" src="https://user-images.githubusercontent.com/5768296/40315499-ef1c2d8e-5cf1-11e8-9585-a985fb29e1e2.png">

bent0b0x:
A fix is in progress via #626

felipegaucho:
Fixed already.

PR #626, opened 2018-05-26
589 player time fix

bent0b0x:
⚠️ perhaps do not merge this yet as I haven't been able to test it in the browser, since the dev env is broken at the moment ⚠️ 

**Issue**
#589 

**Work Done**
- Enforce that the rendered current video time does not exceed the video duration. Due to rounding errors, it is possible that this number unexpectedly exceeds duration, so we needed to account for this.

### DIFF
--- a/src/scripts/selectors/PlayerSelectors.js
+++ b/src/scripts/selectors/PlayerSelectors.js
@@ -33,19 +33,6 @@ export const getPlayingVideo: (
   }
 )
 
-export const getFormattedCurrentTime: (
-  state: RootState
-) => string = createSelector(
-  [getPlayerCurrentTimeSeconds],
-  (currentTimeSeconds: number): string => {
-    const roundedTime: number = Math.floor(currentTimeSeconds)
-    return TimeFormat.fromS(
-      roundedTime,
-      roundedTime >= 3600 ? 'hh:mm:ss' : 'mm:ss'
-    )
-  }
-)
-
 export const getFormattedDuration: (
   state: RootState
 ) => string = createSelector(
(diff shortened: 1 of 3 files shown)


src/scripts/selectors/PlayerSelectors.js
```diff
@@ -33,26 +33,30 @@ export const getPlayingVideo: (
...
+export const getFormattedDuration: (
+  state: RootState
+) => string = createSelector(
+  [getPlayingVideo],
+  (playingVideo: ?VideoRecord): string => {
+    const duration: ?string = playingVideo && playingVideo.get('duration')
+    return formatDuration(duration)
+  }
+)
+
```
bent0b0x: Moved this function up in scope so that `getFormattedCurrentTime` could access it

src/scripts/selectors/PlayerSelectors.js
```diff
@@ -33,26 +33,30 @@ export const getPlayingVideo: (
...
-export const getFormattedDuration: (
-  state: RootState
-) => string = createSelector(
-  [getPlayingVideo],
-  (playingVideo: ?VideoRecord): string => {
-    const duration: ?string = playingVideo && playingVideo.get('duration')
-    return formatDuration(duration)
+    return formattedDuration < formattedCurrentTime
+      ? formattedDuration
+      : formattedCurrentTime
```
bent0b0x: Return the duration if duration is less than what we have calculated as the current time

jellegerbrandy:
ipfs is up again;  let me know when you think it is ready to merge

bent0b0x:
@jellegerbrandy looks good to merge 👍 !

PR #641, opened 2018-05-29
add selectVideoToPublish on click

eliawk:


src/scripts/actions/UploaderActions.js
```diff
@@ -71,7 +71,7 @@ export const uploadAndTranscode = (file: Object, videoId: string) => (
     videoId = paratii.eth.vids.makeId()
   }
   dispatch(videoFetchSuccess({ id: videoId, owner: paratii.eth.getAccount() }))
-  dispatch(selectUploaderVideo(videoId))
+  // dispatch(selectVideoToPublish(videoId))
```
bent0b0x: Let's delete if unused 😄

src/scripts/components/UploadListItem.js
```diff
@@ -289,12 +290,14 @@ class UploadListItem extends Component<Props, Object> {
 
   onPublishVideo (e: Object) {
     e.preventDefault()
+    const videoId = this.state.id
```
bent0b0x: If we have the `video` in props do we need it in `state`? 

If we want the id can we just check `this.props.video.get('id')`?

eliawk: which is the problem to have also the `video id` in the `state`?

bent0b0x: Two issues:

It is duplication of data. If we have it in two places which is our source of truth (it should be props) and how do we guarantee they don’t get out of sync?

Setting state in addition to props can easily lead us to unnecessary rerenders of the component, which is bad for performance. We should avoid putting things in state that are already in props because it is unnecessary and avoids these problems

PR #651, opened 2018-05-31
Fix/uploader stake

eliawk:


src/scripts/components/UploadListItem.js
```diff
@@ -532,13 +546,16 @@ class UploadListItem extends Component<Props, Object> {
...
-                Please be sure not to violate others’ copyright or privacy
-                rights. Learn more
-              </Text>
-              <Text hidden>{urlForSharing}</Text>
+              {!isPublished ? (
+                <Text gray small>
+                  By clicking on the “Publish” button you acknowledge that you
+                  agree to Paratii’s Terms of Service and Community Guidelines.
+                  Please be sure not to violate others’ copyright or privacy
+                  rights. Learn more
```
bent0b0x: Why do we say `Learn more` here? This doesn't look like a clickable link

eliawk: ahah you are right, i don't know i think i was kind a placeholder text..

Issue #561, opened 2018-05-08
Add author name to search results


bent0b0x:
When #530 is resolved, we should display the author name for videos in the search results:

![image](https://user-images.githubusercontent.com/7697924/39735765-0a620322-524b-11e8-897e-8031c06227f3.png)

[Link to design](https://xd.adobe.com/spec/6e9bbad0-7d91-41e4-bc51-a1f31ff57882/#screen/81e820ab-bbb3-42d9-8b25-f4c30ccfa23f/search-future-needs1920)

This is replacing #539

bent0b0x:
I think we've been thinking about the video authorname incorrectly actually. We should not attach an explicit `author` field to the videos at upload or publish time, because I'm sure we will enable people to change their usernames and then we'd have stale `author` fields.

@geckoslair, can we make it so that the db will not only include `owner` address in video query results, but also the db will fetch and attach `ownerUsername` (or `author`) to resulting JSON as well? I think we should do this on all calls to `/videos`, whether it is for search or just fetching one video.

Let me know your thoughts!

jellegerbrandy:
We'd already decided to remove the author field from all the forms, and I think we've done that (looking at staging). 

This basically means that new videos will not get a separate "author" field (they'll just get the name fo the uploader - note that this is not perfect either and we may want to change that again in the future)

I'm not sure what to do with the "old" author names in the existing videos. One of 3 options:

1) keep the legacy, and use a scheme such as ben proposes above
2) remove the authro field from everything, and just loose the existing author names
3) remove the author field from everything, but create some user accounts in which we copy the author field (if we have collisions, we will just choose one).

jellegerbrandy:
personally, I'd be for 2, at most 3.

jellegerbrandy:
wait, I see this is already discussed in https://github.com/Paratii-Video/paratii-db/issues/134, so we continue it there..

geckoslair:
closed at paratii-db side

bent0b0x:
Working on this 👍

PR #652, opened 2018-06-01
561 - author in search results

bent0b0x:
**Note**
I'm having db issues so wasn't able to test this locally, but either way if/when this is working on the db side it will work in the UI.

**Issue**
#561 

**Work Done**
- render video author in search results

**Screenshot**

image

### DIFF
--- a/src/scripts/components/widgets/SearchResult.js
+++ b/src/scripts/components/widgets/SearchResult.js
@@ -9,17 +9,13 @@ import { getVideoThumbnailUrl, getVideoPlayUrl } from 'utils/UrlUtils'
 import { formatDuration } from 'utils/VideoUtils'
 import TruncatedText from 'components/foundations/TruncatedText'
 
-const Wrapper = styled.div`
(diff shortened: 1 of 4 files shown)


bent0b0x:
@pedrocasa we still don't show the author name on the `/play` page. Thoughts on where it should go?

image

src/scripts/components/widgets/SearchResult.js
```diff
@@ -114,33 +116,34 @@ class SearchResult extends React.Component<Props, void> {
...
+            </Duration>
+          </ThumbnailData>
+        </ThumbnailWrapper>
+        <Info>
+          <TopBar>
+            <Title data-test-id="search-result-title">
+              <TruncatedText>{video.get('title')}</TruncatedText>
+            </Title>
+            <Author data-test-id="search-result-author">
+              <TruncatedText>{video.get('author')}</TruncatedText>
```
bent0b0x: We truncate the author name if it is too long

PR #659, opened 2018-06-01
[WIP] Fix/dependency is expression

jellegerbrandy:


.circleci/config.yml
```diff
@@ -33,10 +33,10 @@ jobs:
...
           command: |
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
-              yarn build:staging
+              yarn build:production
               ./deploy.sh production
             fi
             if [ "${CIRCLE_BRANCH}" == "dev" ]; then
-              yarn build
+              yarn build:staging
               ./deploy.sh staging
```
bent0b0x: 😮 good fix

jellegerbrandy:
unfortunatly, this approach does not work, because it uses `fs` to read the files (so we did not need to require them), which the browser does not like.

Issue #524, opened 2018-05-01
Write test for search


bent0b0x:
- [ ] unit tests
- [x] functional tests
- [x] functional tests for "load more results" functionality

PR #663, opened 2018-06-03
524 - add test for more search results

bent0b0x:
**Issue**
#524 

**Work Done**
- add chimp tests for "load more search results" functionality
- do a little bit of test refactoring

### DIFF
--- a/src/scripts/components/pages/SearchResults.js
+++ b/src/scripts/components/pages/SearchResults.js
@@ -56,7 +56,7 @@ const LoaderWrapper = styled.div`
   align-items: center;
 `
 
-const HasNextLink = styled.button`
+const HasNextButton = styled.button`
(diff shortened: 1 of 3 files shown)


src/scripts/components/pages/SearchResults.js
```diff
@@ -85,9 +85,12 @@ class SearchResults extends React.Component<Props, void> {
...
 
     return (
-      <HasNextLink onClick={searchForMoreVideos}>
+      <HasNextButton
+        data-test-id="more-results-button"
+        onClick={searchForMoreVideos}
+      >
         Click for more results
-      </HasNextLink>
+      </HasNextButton>
```
bent0b0x: Renamed this as `Button` makes more sense than `Link` in this context

Issue #661, opened 2018-06-03
De-dup sass styles in portal app and embedded player


bent0b0x:
We currently have some `scss` written for the embedded player in [embed/index.scss](https://github.com/Paratii-Video/paratii-portal/blob/dev/src/styles/embed/index.scss) and similar (if not nearly entirely identical?) styles for the rest of the portal app written in [app.scss](https://github.com/Paratii-Video/paratii-portal/blob/dev/src/styles/app.scss). The styles for both should be almost entirely the same, except for some small exceptions. We should de-dup these to avoid bugs in the future.

We can probably inject these styles via a template on the server-side.

PR #664, opened 2018-06-03
661  - dedup styles

bent0b0x:
**Issue**
#661 

**Work Done**
- delete unused server route
- remove duplicate styles between portal and embedded app

**Note**
This new approach could be a concern in the future (and we can implement a different approach) if the portal someday needs a _ton_ more `.scss` styles that are unnecessary in the embedded app, or vice versa. But I do not see this happening anytime soon, and this solution prevents us from accidentally changing a shared style in one app and not the other


src/server/routes/embed.js
```diff
@@ -1,113 +0,0 @@
-/* @flow */
```
bent0b0x: This was not used

src/styles/embed/index.scss
```diff
@@ -1,37 +1,13 @@
-@import '../reset.scss';
+@import '../app.scss';
```
bent0b0x: Now we import the portal app styles into the embedded stylesheet

Issue #452, opened 2018-04-10
Do not allow users to select video file types that we do not support


bent0b0x:
Right now the user can select any video file (and maybe even images?) to upload. However, if they select something other than an `.mp4` the upload works but transcoding will hang forever and then nothing happens.

We can restrict on the client the type of files the user can select.

@pedrocasa @felipegaucho thoughts on adding some verbiage in the uploader to tell the user about this restriction?

pedrocasa:
Best way to solve it: don't let the user go to the upload page. If he tryes to upload an jpeg(ie) a notification should appear saying " invalid format" and he would stay on the landing page

pedrocasa:
I totally agree with u

PR #667, opened 2018-06-03
[WIP] 452 - restrict uploads to mp4 (and add some messaging)

bent0b0x:
**Issue**
#452 

**Work Done**
- restrict file input to only accept `.mp4` by default
- add messaging to indicate that only `.mp4`s are accepted

**Known limitations**
- users can still drag non-mp4 files into the dropzone. But at least we are warning them now that this will fail. Detecting the file type of a dragged file is possible but not sure if it is worth the time right now. And it would likely implore us a dependency which I would like to avoid until we get the bundle to a more manageable size. In addition, I believe it would require us to convert the entire file into a `Uint8Array` (which would be a performance concern depending on file size).

**Screenshots**

image

image

other file types are not selectable
image


bent0b0x:
@ya7ya is mp4 still the only supported file type?

bent0b0x:
@pedrocasa @felipegaucho what do you think? I'm happy to update the messaging, both the words and where they are located

src/scripts/components/Dropzone.js
```diff
@@ -1,657 +0,0 @@
...
-import React from 'react'
-import PropTypes from 'prop-types'
-import styled from 'styled-components'
-
-const DropzoneWrapper = styled.div`
-  cursor: pointer;
-  width: 200px;
-  height: 200px;
-  border: 2px dashed #ccc;
-  background: #666;
```
bent0b0x: Unused file 😄

src/scripts/components/widgets/FilesUploader.js
```diff
@@ -147,9 +146,36 @@ class FilesUploader extends Component<Props, Object> {
...
+          onClick={this.onCheck}
+          onChange={this.onFileChosen}
+          onDragEnd={this.onDrag}
+        />
+        <UploadCover>
+          {card && <Title>Upload video</Title>}
+          <UploadCoverIcon>
+            {card && (
+              <Icon>
+                <FilesUploaderSvg />
```
bent0b0x: Refactored some duplicate code

pedrocasa:
Amazing work Ben, maybe I would change the color and add the (+) icon

bent0b0x:
Sounds good. @pedrocasa what color? I'll leave the plus icon out for this one as that is being handled by a separate issue

pedrocasa:
let's use this #65667C

jellegerbrandy:
marked this pr as [wip] so we won't merge it accidentally, let me know when it's done

Issue #668, opened 2018-06-03
Securing wallet when trying to upload a video requires redundant action


bent0b0x:
**Steps to  reproduce**
- navigate to `/upload`
- click upload button (or drag file into dropzone)
- enter password

**Expected**
- file picker appears once you have secured wallet

**Actual**
- nothing happens after wallet is secured and you need to click again

![securefail](https://user-images.githubusercontent.com/7697924/40891340-dafa0b5e-6751-11e8-9fa0-dbe18f2fb48f.gif)

PR #669, opened 2018-06-03
668 - improve UX securing/creating wallet when trying to upload

bent0b0x:
**Issue**
#668 

**Work Done**
- When you first access the `/upload` route, request password
- If user secures wallet, then go and fetch their videos
- **when user cancels out of secure flow, redirect them to homepage** 👈 @jellegerbrandy this is new behavior that I kind of decided on (but happy to reevaluate it). I think it makes sense. If you don't secure your account you have no business on the upload page (in fact we won't let you do anything anyway). LMK your thoughts

**Demos**

Successful login
uploaduxwin

Cancels password modal
passwordclosewin

Cancels recovery modal
recoveryclosewin


src/scripts/components/widgets/modals/Modal.js
```diff
@@ -163,26 +168,42 @@ class Modal extends Component<Props, void> {
...
+        ModalComponent = ModalRewriteSeed
+        break
       case MODAL.RESTORE_ACCOUNT:
-        return <ModalRestoreAccount />
+        ModalComponent = ModalRestoreAccount
+        break
+    }
+
+    if (ModalComponent) {
+      return <ModalComponent {...modalProps} />
```
bent0b0x: Now we can dispatch modal actions that include addl properties to pass to the modal component

src/scripts/components/widgets/modals/ModalAskPassword.js
```diff
@@ -76,37 +82,41 @@ class ModalAskPassword extends Component<Props, Object> {
...
-      this.props.notification(
-        {
-          title: 'The password is not valid',
-          message: 'Please retype the password.'
-        },
-        'error'
-      )
-    }
+
+    setTimeout(() => {
```
bent0b0x: This timeout is necessary to show the notification "Trying to unlock your keystore...". We are doing this elsewhere in the app

src/scripts/containers/FileUploaderContainer.js
```diff
@@ -3,7 +3,7 @@ import { bindActionCreators } from 'redux'
 
 import { uploadAndTranscode } from 'actions/UploaderActions'
 import type { RootState } from 'types/ApplicationTypes'
-import FilesUploader from '../components/widgets/FilesUploader'
+import FileUploader from '../components/widgets/FileUploader'
```
bent0b0x: Renamed file to be consistent w/ the container

src/scripts/containers/widgets/modals/ModalContainer.js
```diff
@@ -1,13 +1,14 @@
...
 import { bindActionCreators } from 'redux'
-import { getModalContent, getModalStatus } from 'selectors/index'
+import { getModalName, getModalProps, getModalStatus } from 'selectors/index'
 import { closeModal } from 'actions/ModalActions'
 import Modal from 'components/widgets/modals/Modal'
 import type { RootState } from 'types/ApplicationTypes'
 
 const mapStateToProps = (state: RootState) => ({
-  modalContent: getModalContent(state),
+  modalName: getModalName(state),
```
bent0b0x: Changed to `modalName` as it is a better representation of what this property is

src/scripts/reducers/ModalReducer.js
```diff
@@ -4,19 +4,21 @@ import ModalRecord from 'records/ModalRecord'
...
-      previousModalContent: state.modalContent,
-      showModal: true
+  [OPEN_MODAL]: (state: ModalRecord, action: Action<ModalPayload>) => {
+    return state.withMutations((mutableState: ModalRecord) => {
+      mutableState.merge({
+        name: action.payload.modal,
+        previousModalName: state.get('name'),
+        showModal: true
+      })
+      mutableState.set('modalProps', action.payload.modalProps)
```
bent0b0x: If we include `modalProps` in `merge` then they will automatically be converted into an immutable record. This is a case where we don't need/really want to convert something to an immutable in our store.

src/scripts/components/widgets/modals/ModalSetPin.js
```diff
@@ -1,169 +0,0 @@
-/* @flow */
-import React, { Component } from 'react'
```
bent0b0x: Unused modal

src/scripts/containers/widgets/modals/ModalSetPinContainer.js
```diff
@@ -1,19 +0,0 @@
-/* @flow */
-
-import { connect } from 'react-redux'
-import { bindActionCreators } from 'redux'
-import { openModal, closeModal } from 'actions/ModalActions'
```
bent0b0x: Unused container

jellegerbrandy:
About the login moment: my opinion is that is should happen earlier: at the moment the user visits the "uploads" page. Two reasons:
 (1) it is simpler and more standard and
(2) not seeing your videos on the uploads page is confusing (typical case is: go to the page, see no vids, upload one vid, login during the process, return to the uploads page, see the new vid together with your older vids, be confused why you seem them now and not before.

@bent0b0x , if you have no objections, let's do it like this.

bent0b0x:
@jellegerbrandy makes sense, I'll refactor for that UX!

jellegerbrandy:
@bent0b0x , conflicts arose (sorry for the delay in merging), can you resolve them?

bent0b0x:
@jellegerbrandy fixed conflicts. Not sure if tests will pass on circle yet but i'll fix that later.

There is one thing we need to iron out, though. This new flow is that we show the password prompt upon landing on the upload page. On the contrary, on the landing page we need to show it when you attempt to upload the file (obviously can't ask for password on landing page). See #721. So we're going to have two different authorization behaviors on different pages for essentially the same user-driven action. Is this ok?

jellegerbrandy:
i don think it is very useful to think in terms of "flows" here. When to
authenticate is better described as attributes of actions. I.e.
we have pages and actions for which you need to be authenticated, and
pages/action for which you are not.

"Uploading" needed authentication, and that has not changed. Now, also the
'upload page' needs authentication, which seems correct (bc it needs to
show **your** uploads). I see no "different behaviors for essentially the
same action" here...

tldr: It's ok :-)

PR #677, opened 2018-06-05
Ownership proof field

eliawk:


src/scripts/components/UploadListItem.js
```diff
@@ -247,8 +245,9 @@ class UploadListItem extends Component<Props, Object> {
     this.state = {
       open: false,
       id: theVideo.id,
-      title: theVideo.title,
-      description: theVideo.description,
+      title: theVideo.title || '',
+      description: theVideo.description || '',
+      ownershipProof: theVideo.ownershipProof || '',
```
bent0b0x: Can we please not put these in `state`? What is the point? They are already in `props`. The duplication is going to lead to confusion down the line. If we need something from the video we should only access it from `this.props.video`

eliawk: @bent0b0x the props just set the initial state, then when the user change the input field we need to update the state. is there another solution for that?

bent0b0x: Ah ignore me 😊

There are other potential solutions, such as putting the edited video in redux too. That has some benefits but is not necessary for us now.

PR #680, opened 2018-06-05
New account should be saved  #694

jellegerbrandy:
test for #694


eliawk:
this is always null!
```
const newAddress = await getAccountFromBrowser()
 // FIXME this newAddress is null!!!!!
```

PR #681, opened 2018-06-05
[wip] integration tests

jellegerbrandy:


bent0b0x:
I suggest we add `paratii-db` as a `devDependency` of `paratii-portal` so that we are guaranteed to have it when testing. Then the tests scripts can know that the db is available and not rely on something external downloading/copying it

Issue #684, opened 2018-06-06
create Tipping UI


pedrocasa:
Here are the designs of the tipping process. We have designed a flow that gonna be used in the embeddable player and inside the portal. 

- Hardcoded values: No room for arbitrary amounts of PTI.

- We talked about where to show the errors and rather show them all in the same place (in the password confirmation page), no notifications or extra screens. See if the flow looks good to you and if it's fine to fetch the username at the moments we propose.

- The Thank you screen has three options (two that need the name of the user). If we can have them randomly appearing would be awesome. Every tip a new experience.

- Talking about experience, it would be cool to have a coin-like SFX when the TX is confirmed (e.g.: https://www.youtube.com/watch?v=cQ5JIcDx3fg - if it's too kitschy, just ignore). 

- The U$ value is just a placeholder and shouldn't be implemented now

![tip_flow](https://user-images.githubusercontent.com/30123644/41008954-6ce12cf6-6904-11e8-9648-0e7f2e24a66b.png)

Adobe prototype link
https://xd.adobe.com/view/c34704c3-94b6-4504-aac2-44e2c19c8ced

- [ ] Add tests!

bent0b0x:
I love it! For the thank you screen I am a fan of the second option

felipegaucho:
Could also have a small tweet button on the thank you screen (tweets the video + "I just tipped + the thank you sentence".

eliawk:
@pedrocasa super nice!

jellegerbrandy:
Very cool. Not sure about the kitchy kedching effect :-)

(ben: if I understand pedro correctly, the idea is not that we choose a
thank you screen, but we show on of these three randomly...)

pedrocasa:
@bent0b0x  here are the specs link and the assets to develop it.

specs https://xd.adobe.com/spec/a4036521-bcd4-454b-8f9d-d0093b02c4d6
assets https://we.tl/U84fDeZ1Nj

bent0b0x:
@felipegaucho @pedrocasa when do we show the "tip" prompt/button? Right when the video starts? I think that seems spammy and maybe it's good to wait a little?

Let me know your thoughts!

bent0b0x:
Another question: where has this symbol come from?

![image](https://user-images.githubusercontent.com/7697924/41326672-0b7eb2a0-6e8e-11e8-8c0a-67b68c1fd2e5.png)

Is this a standard way of representing USD? It is not one I have seen

[...]

PR #705, opened 2018-06-13
684 - Tipping

bent0b0x:
**Issue**
#684 

**Work Done**
- add some basic components to start on tipping UI

**TODO**
- [x] get tipping to actually work 😄 
- [x] show tipping button after 3s even if user scans to a time before 3s
- [x] in addition to showing an error when you attempt to transfer a PTI balance that you do not have, we should disable tipping altogether if we detect you do not have enough PTI and disable tipping options that are above your balance
- [x] convert PTI `toWei` before sending to `paratii.eth.transfer`
- [ ] get `paratii.eth.transfer()` to work

**NOTES**
- The tipping experience is only for the portal. Tipping requires an account, and we have not worked on the login/secure flow in the embeddable player
- You might notice some of the tipping copy's are incorrect/weird. This is because I was testing w/ videos that lacked an attached `username`. This will not happen for any videos that are created correctly.

**Demo**

tipwin

tipsecurefirstwin

tippasswordfailwin

**When some tip amounts are disabled because you do not have enough PTI**
image


src/scripts/reducers/PlayerReducer.js
```diff
@@ -35,7 +38,24 @@ const reducer = {
...
+    )
+
+    return state.merge({
+      currentTimeSeconds: action.payload.time,
+      latestPlaybackTimestamp: timestamp,
+      totalTimeViewedSeconds: prevPlaybackTimestamp
+        ? prevTotalTimeViewedSeconds +
+          (timestamp - prevPlaybackTimestamp) / 1000
+        : 0
+    })
```
bent0b0x: These changes make it so that we keep track of the total time spent playing the video, regardless of which specific times in the video you have seen. This way we can say "you have been watching this video for 3 seconds"

jellegerbrandy:
Very nice. Some remarks:

- I have 0 balance, but tipping 15 PTI went through the UX without problems, including confirmation.  
- If I'm not logged in, I get TWO password prompts, first wehn clicking on the "give a tip" button, and then later to confirm the amount. One is enough, I'd say (and I'd just remove the first one)
- Can we put a placeholder if the name of the user is not known? "Anon is very grateful" is still better than "is very grateful" :-)

bent0b0x:
@jellegerbrandy

- will look into balance requirements, that’s defjnitelt not supposed to happen
- that’s also not supposed to happen. If you have secured in the last 60 seconds, the password prompt during tipping should not appear
- yes, I think that is a good idea

bent0b0x:
>I have 0 balance, but tipping 15 PTI went through the UX without problems, including confirmation.

@jellegerbrandy I've created [this issue](https://github.com/Paratii-Video/paratii-js/issues/285) in `paratii-js`. I don't see that `paratii.eth.transfer` throws an error letting us know that a transfer failed due to insufficient funds

bent0b0x:
>If I'm not logged in, I get TWO password prompts, first wehn clicking on the "give a tip" button, and then later to confirm the amount. One is enough, I'd say (and I'd just remove the first one)

@jellegerbrandy resolved by 11a9cd1c35aabeb94d8b90620a846cfc7d7aa67f

bent0b0x:
>Can we put a placeholder if the name of the user is not known? "Anon is very grateful" is still better than "is very grateful" :-)

@jellegerbrandy fixed via eee024b. Added a fallback author name of "the author" (which will be translated)

bent0b0x:
@jellegerbrandy I think this PR is ready (please do give it another test and lmk what you think).

Despite the issue highlighted by [this comment](https://github.com/Paratii-Video/paratii-portal/pull/705#issuecomment-398949512), I think we're good. The latter issue is about the fact that `paratii-js` does not throw an error if you lack the sufficient balance for a particular transfer. However, the UI has guards against this now. Namely:
- you cannot select a tip amount greater than your balance
- if you don't have enough PTI for any of the tip amounts, the tip button simply never shows

src/scripts/components/tipping/TipOverlay.js
```diff
@@ -0,0 +1,176 @@
...
+      currentStep: TIPPING_UI_STEPS.CHOOSE_AMOUNT
+    })
+  }
+
+  transferTip = async () => {
+    try {
+      const tipAmountWei: number = Number(
+        ParatiiLib.eth.web3.utils.toWei(`${this.state.tipAmount}`)
+      )
+      await ParatiiLib.eth.transfer(
```
bent0b0x: @jellegerbrandy I just added a `toWei` conversion above, but I'm still seemingly able to transfer any amount to anyone. Going to debug some more.

pedrocasa:
do we need the notification boxes in this case

pedrocasa:
Maybe a better way to show if there's not enough PTI should be this:

40% of opacity - "unclickable" elements







I didn't design it on my flow because I did not know if it was possible to implement in that way.

bent0b0x:
@jellegerbrandy does this branch work locally for you now? I am simply unable to get `.transfer` to work whether it is in the portal or the command line. However if transfers work on your machine (and others) then the logic is in there such that this functionality is fine. 

However, I am not sure if the issues w/ `.transfer` that I am seeing are systemic and more widespread than just me

bent0b0x:
https://github.com/Paratii-Video/paratii-portal/issues/757 is resolved by 25f4a62

Issue #699, opened 2018-06-12
sign up/ login new button


pedrocasa:
At the moment our login and signup button are not so appealing. In order to change it, I've designed a new version of it

![login-button](https://user-images.githubusercontent.com/30123644/41311640-3586f00e-6e5b-11e8-92d9-4ac95ef63265.png)

![login-signup](https://user-images.githubusercontent.com/30123644/41311642-3655756e-6e5b-11e8-8bd5-9face0fe8057.png)

- the first header will only appear if the account is already stored on the browser.
- the second will appear if there's not an account registered on the browser.

dropboxlink: https://www.dropbox.com/home/paratii-design/master
specs link: https://xd.adobe.com/spec/d218de3e-ce11-4bfa-a41e-d90ce992644f

few aesthetic changes that we may change on the header in order to look better.

- round the corners of the search box
- decrease the height of the header to 62 pixels (search box 38)
- the color of the header should be: blackAlt
- the color of the buttons and the graphic elements inside the search box (word: search and the 
magnifying glass icon) header should be: gray

pedrocasa:
@jrgarou 
In the current staging and portal when your account is not stored on the browser, there's a button with only the word sign up.

![screen shot 2018-06-29 at 11 09 54 am](https://user-images.githubusercontent.com/30123644/42097039-62b6a544-7b8d-11e8-8194-40531ce8c40f.png)


We need the word log in as well for users that already have an account

PR #730, opened 2018-06-18
699 - new header

jrgarou:
#699 

image

### DIFF
--- a/src/scripts/components/foundations/Button.js
+++ b/src/scripts/components/foundations/Button.js
@@ -1,5 +1,32 @@
 import styled, { css } from 'styled-components'
 
+export const ButtonStyleBackground = css`
+  ${props => {
+    let _background: String
+
+    if (props.white) {
+      _background = props.theme.colors.button.white
+    } else if (props.gray) {
+      _background = props.theme.colors.button.gray
+    } else if (props.pink) {
+      _background = props.theme.colors.button.pink
+    } else if (props.green) {
+      _background = props.theme.colors.button.green
+    } else {
+      _background =
+        'linear-gradient(135deg, ' +
+        props.theme.colors.VideoForm.info.progress.barFrom +
+        ' 0%, ' +
+        props.theme.colors.VideoForm.info.progress.barTo +
+        ' 100%)'
+    }
+
+    return css`
+      background: ${_background};
+`
+  }};
+`
+
 export const ButtonStyleColor = css`
   ${props => {
     let _color: String
@@ -10,6 +37,8 @@ export const ButtonStyleColor = css`
       _color = props.theme.colors.button.purple
     } else if (props.pink) {
       _color = props.theme.colors.button.pink
+    } else if (props.green) {
+      _color = props.theme.colors.button.green
     } else {
       _color = props.theme.colors.button.gray
     }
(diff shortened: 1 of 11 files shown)


src/scripts/components/foundations/Button.js
```diff
@@ -40,16 +69,38 @@ export const StyleAnchor = css`
...
+    props.anchor
+      ? props.theme.fonts.weight.regular
+      : props.theme.fonts.weight.bold};
+  margin: ${props => props.margin};
+  opacity: ${props => (props.disabled ? 0.5 : null)};
+  pointer-events: ${props => (props.disabled ? 'none' : null)};
+  text-transform: ${props => (props.anchor ? null : 'uppercase')};
+  user-select: ${props => (props.anchor ? null : 'none')};
+  text-rendering: geometricPrecision;
+`
```
bent0b0x: Can we move these to another file? Is this a good PR to refactor this component to be more generic? Fine if not 😄

src/scripts/components/foundations/Button.js
```diff
@@ -40,16 +69,38 @@ export const StyleAnchor = css`
...
+    filter: brightness(1.3);
   }
 `
 
+export const ButtonColor = styled.button`
+  ${ButtonStyle} ${ButtonStyleHover} ${ButtonStyleBackground} border-radius: 2px;
+  color: ${props => props.theme.colors.button.white};
+  padding: 8px 18px;
+`
+
```
bent0b0x: Also, in general I think we may want to examine this component a bit. I tried to use it for tipping but found some bugs w/ the active/hover states that I could not easily work around. Essentially the button appeared in those states inappropriately after clicking on and off. 

Just an FYI (have you seen issues like this?)

src/scripts/components/foundations/forms/Input.js
```diff
@@ -9,10 +9,16 @@ const AutofillFix = props => keyframes`
...
+  color: ${props => props.theme.colors.TextField.color};
   font-family: ${props => props.theme.fonts.family}, sans-serif;
   font-size: ${props => props.theme.fonts.form.input};
   font-weight: ${props => props.theme.fonts.weight.regular};
   width: 100%;
+
+  &:-webkit-autofill {
+    animation-name: ${AutofillFix};
+    animation-fill-mode: both;
+  }
```
bent0b0x: What about other browsers?

bent0b0x: Ah I see this was just moved. But what about other browsers?

src/scripts/components/structures/header/MainNavigation.js
```diff
@@ -106,7 +98,7 @@ class MainNavigation extends Component<Props, Object> {
 
           {!this.props.isWalletSecured ? (
             <NavItem>
-              <NavLinkPurple
+              <ButtonColor
```
bent0b0x: The name `ButtonColor` seems confusing. That should be a color variable, I would think. What about `<ColoredButton />`?

src/scripts/components/structures/header/MainNavigation.js
```diff
@@ -4,7 +4,7 @@ import React, { Component } from 'react'
 import { Link } from 'react-router-dom'
 import styled from 'styled-components'
 
-import Button from 'components/foundations/Button'
+import Button, { ButtonColor } from 'components/foundations/Button'
```
bent0b0x: Can we move `ButtonColor` to another file?

jellegerbrandy:
ok @jrgarou , can you address ben's comments? Then feel free to merge :-)

src/scripts/components/foundations/Button.js
jrgarou: @bent0b0x let's create an issue for all those changes? 
What I think about: 
- Button(colored button)
- TextButton(actually Button)

And share(`export`) its styles.

src/scripts/components/foundations/Button.js
jrgarou: >have you seen issues like this?

Probably some `:active` selector. Could you send more info about this, please?

src/scripts/components/foundations/forms/Input.js
jrgarou: this is only for the yellow background from webkit

src/scripts/components/structures/header/MainNavigation.js
jrgarou: Could we do those changes with **[this](https://github.com/Paratii-Video/paratii-portal/pull/730#discussion_r196427367)** ?

src/scripts/components/structures/header/MainNavigation.js
jrgarou: Could we do those changes with **[this](https://github.com/Paratii-Video/paratii-portal/pull/730#discussion_r196427367)** ?

src/scripts/components/foundations/Button.js
bent0b0x: I think we should definitely create an issue to discuss further. 

Not sure about that distinction, though, and my initial thought is that we should export any styles from components. I believe there are other ways that we can share styles as needed, via constants and such. Component files should only be components, and should not have any other responsibilities like maintaining styles for other components (unless a component is a direct extension of said component). 

I also feel like if we need multiple components to share many of the same styles then perhaps we have not written our abstractions correctly, and we can refactor a bit

bent0b0x: https://github.com/Paratii-Video/paratii-portal/issues/737

PR #735, opened 2018-06-19
732 landing page links

jrgarou:
I did replace the anchor tag for `Link` router

I don't think it need to create a conditional to replace the Link for anchor on `process.env.NODE_ENV != production`. What do you thinik, guys?


jellegerbrandy:
I don't think so either (I asumme the links will be broken in dev, but that's not a real problem)

PR #750, opened 2018-06-29
Disable ModalStake button when staking

jrgarou:
https://github.com/Paratii-Video/paratii-portal/blob/ae526fd800618301c9a8dc0f67cb105d5b647fbc/src/scripts/components/widgets/modals/ModalStake.js#L187

### DIFF
--- a/src/scripts/components/widgets/modals/ModalStake.js
+++ b/src/scripts/components/widgets/modals/ModalStake.js
@@ -52,7 +52,8 @@ class ModalStake extends Component<Props, Object> {
     this.state = {
       errorMessage: false,
       agreedTOC: false, // TODO,
-      stakeAmount: 0
+      stakeAmount: 0,
+      disableButton: false
     }
     this.onSubmit = this.onSubmit.bind(this)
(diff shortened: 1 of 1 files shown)


src/scripts/components/widgets/modals/ModalStake.js
```diff
@@ -65,6 +66,10 @@ class ModalStake extends Component<Props, Object> {
     const stakeAmountWei = paratii.eth.web3.utils.toWei(String(stakeAmount))
     const videoIdStaked = this.props.selectedVideo.id
 
+    this.setState({
+      disableButton: true
```
bent0b0x: I think _preferably_ we would store `this.state.isStaking` instead of `disableButton` directly. This way we are storing, in `state`, the actual thing that we are tracking.

Then we can have a method in the component like so:

```javascript
disableButton = () => {
   return this.state.isStaking;
}
```

Seems a bit over-engineered, but is better, I think, in case we ever have to add other conditions as to when the button needs to be disabled

PR #753, opened 2018-06-30
Do not reload balances on each render

bent0b0x:
Calling a function that will update state/props of a component can easily lead to a ton of re-renders across many components. We should avoid this at all costs 😅

### DIFF
--- a/src/scripts/components/UserNav.js
+++ b/src/scripts/components/UserNav.js
@@ -162,17 +162,20 @@ const UserSubNavListItemLink = styled(Link)`
 class UserNav extends Component<Props, Object> {
   UserNavWrapper: HTMLElement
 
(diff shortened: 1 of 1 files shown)


src/scripts/components/UserNav.js
```diff
@@ -162,17 +162,20 @@ const UserSubNavListItemLink = styled(Link)`
 class UserNav extends Component<Props, Object> {
   UserNavWrapper: HTMLElement
 
+  constructor (props: Props) {
+    super(props)
+
+    props.loadBalances()
+  }
```
bent0b0x: If we want to refresh the balances more, we should intentionally do it on some interval. We cannot simply do it on each `render`

Issue #763, opened 2018-07-04
After making a stake, the wallet values should be updated


jellegerbrandy:
I.e. both the values of "how much PTI do you have" as well as the "total staked" should be updated

Issue #756, opened 2018-07-03
Balance needs to be updated


jellegerbrandy:
We used to do polling here to update the balance. I'm not clear on what the status of that is, but apparently it is not working :-)

reproduce in dev:

* run the portal, log in, look at your balance 
* run `node dev/faucet.js` to send yourself some PTI
* see that your balacne is not updated [<- this is not good]
* reload the page, login again, see updated balance

eliawk:
@jelle after the faucet my balance has been updated..so it seems working!

PR #770, opened 2018-07-08
756 - regularly update account balance and update total staked

bent0b0x:
**Issue**
#756 
#763 

**Work Done**
- load account balance every 2 seconds
- remove polling logic from `<PTIBalance />` component and move the logic to the application itself
- update staking balance after making a stake (this balance is not updated every 2 seconds because I think this addl request is unnecessary. In general we should only have to update this balance when we explicitly make a stake on the FE)

**Demos**

👇 **balance**

balanceupdate

☝️ resolution on the gif may not be great, but the gif is demonstrating that when we update the balance w/ `dev/faucet.js` the balance updates in the app immediately (and at worst, after 2 seconds)

👇 **total staked**

stakewin

### DIFF
--- a/src/scripts/components/widgets/PTIBalance.js
+++ b/src/scripts/components/widgets/PTIBalance.js
@@ -8,8 +8,7 @@ import TruncatedText from 'components/foundations/TruncatedText'
 
 type Props = {
   balance: string,
-  color?: string,
-  loadBalances: () => void
+  color?: string
 }
 
 const Wrapper = styled.div`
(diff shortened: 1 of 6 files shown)


src/scripts/components/widgets/PTIBalance.js
```diff
@@ -23,24 +22,7 @@ const NumberWrapper = styled.span`
...
-  componentWillMount = (): void => {
-    const { loadBalances } = this.props
-    this.intervalId = setInterval(loadBalances, REFRESH_BALANCES_INTERVAL_MS)
-  }
-
-  componentWillUnmount = (): void => {
-    if (this.intervalId) {
-      clearInterval(this.intervalId)
-    }
-  }
```
bent0b0x: This logic was moved

jellegerbrandy:
Great @bent0b0x, thanks!

Issue #779, opened 2018-07-11
Unpublished videos are being published


pedrocasa:
Videos that I did not give stake are being published and going to the page of my videos.

This is happening on both  Staging and Production

![screen shot 2018-07-11 at 10 30 35](https://user-images.githubusercontent.com/30123644/42575124-3623440c-84f6-11e8-9d0f-2a7809e9d4ea.png)

![screen shot 2018-07-11 at 10 30 22](https://user-images.githubusercontent.com/30123644/42575088-2d0e02d0-84f6-11e8-9cce-248141eaeb42.png)

![screen shot 2018-07-11 at 12 17 25](https://user-images.githubusercontent.com/30123644/42582010-7114aa48-8504-11e8-913f-2a4ed3683d05.png)

bent0b0x:
IMO unpublished videos should appear on my videos and we should allow you to publish from there. 

But we should update our logic to not allow playing of unpublished videos

jrgarou:
I've removed the link from unpublished videos and show the correct status - only _publish_ and _unpublish_
The commit/branch: https://github.com/Paratii-Video/paratii-portal/commit/fe324199620dc886f859c6d423422d7a352c2054


> IMO unpublished videos should appear on my videos and we should allow you to publish from there.

There's no button to publish the video, but I think it's a good idea, @bent0b0x 
What you think, @pedrocasa? 


--

![image](https://user-images.githubusercontent.com/320157/42587240-77280270-8510-11e8-8256-88ce6a67de7f.png)

pedrocasa:
OK, 

the solution, for now, use the status as ''published'' or ''unpublished''

If the user clicks on the published ones he will go to the video page and if he clicks on the unpublished ones a modal ( same as the one that appears on the upload process) would appear.
  
![web 1920 22](https://user-images.githubusercontent.com/30123644/42600318-3e857628-8538-11e8-8bb8-f72444af60bb.png)

![web 1920 23](https://user-images.githubusercontent.com/30123644/42600320-4018e722-8538-11e8-8e45-09ec4103c29b.png)



I still think that showing the published and unpublished videos may be difficult for (especially when add more subcategories such as Whitelisted, challenged ... etc). but for now, this solution may handle it

Issue #778, opened 2018-07-10
Implement the edit videos on the published page [my videos]


pedrocasa:
We need a way for the users to edit their videos after publishing it ( change the title/description ...)

![screen shot 2018-07-10 at 17 19 27](https://user-images.githubusercontent.com/30123644/42535344-81553f20-8465-11e8-83d2-596f06a968eb.png)

The idea is that there will be a button on each thumbnail that will take the users to an

![screen shot 2018-07-10 at 16 56 22](https://user-images.githubusercontent.com/30123644/42535713-8468fdfe-8466-11e8-8fed-4fe3ce5711f7.png)
 edit video page 

![web 1920 18](https://user-images.githubusercontent.com/30123644/42535761-ab8c3720-8466-11e8-9252-27610fd00430.png)

PR #780, opened 2018-07-11
779 hide myvideos unpublished

jrgarou:
More details: https://github.com/Paratii-Video/paratii-portal/issues/779

### DIFF
--- a/src/translations/pt-BR.json
+++ b/src/translations/pt-BR.json
@@ -16,6 +16,16 @@
             "description": "Explore curtas, clipes, e novos formatos que nossa comunidade inventa todos os dias."
         }
     },
+    "myVideos": {
+        "title": "My videos",
+        "item": {
+            "status": {
+                "label": "Status",
+                "published": "Publicado",
+                "unpublished": "Não publicado"
+            }
+        }
+    },
     "navigation": {
         "voucher": "Obter PTI",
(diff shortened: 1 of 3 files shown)


src/translations/pt-BR.json
```diff
@@ -16,6 +16,16 @@
...
+    "myVideos": {
+        "title": "My videos",
+        "item": {
+            "status": {
+                "label": "Status",
+                "published": "Publicado",
+                "unpublished": "Não publicado"
+            }
+        }
+    },
```
bent0b0x: (I don't speak Portuguese but) are all of these actually translated?

I think we should leave files other than `en.json` untouched so that they can be fully translated later by Daniel

bent0b0x:
Looks good. Mind adding a before/after screenshot just because I think the solution changed from what was posted in the original issue?

src/translations/pt-BR.json
jrgarou: Ok. If it is fine to Daniel take care of that I'll not replicate the keys

jrgarou:
@bent0b0x, this PR is just for remove the click from unpublished videos and show the publish status.
I'm working on another issue (https://github.com/Paratii-Video/paratii-portal/issues/778). After that, I could finish it.

PR #783, opened 2018-07-12
778 - Edit video page

jrgarou:
I've copy the code from `UploadListItem` to `VideoForm`, to use it in `/upload` and `/edit/VIDEOID` pages.

#### Let's remove the `UploadListItem` and (re) use the  `VideoForm`.

|

image
image


eliawk:
@jrgarou why did you copy the component instead of import it? 
now we have the same code in two different places and this is not very good

jrgarou:
@eliawk It is not two, just the VideoForm. It is a replacement.

PR #822, opened 2026-05-25
feat: add file size restriction messaging for uploads

### LINKED ISSUE #718: File size restriction messaging
We're currently limited to .mp4 files under 800mb as far as uploads go, but the portal doesn't exactly make that clear. We should add two things:

1 - add <800mb to the (only .mp4 currently supported) notice when uploading
2 - add a specific error prompt when you attempt to upload an unsupported file format or size - it's currently a generic "something went wrong with your file, try to retry" message.

Karry2019web:
## Summary

This PR adds clear file size restriction messaging for the uploader, as requested in issue #718.

## Changes

1. **Uploader text**: Updated the "supported file types" notice from `(only .mp4 currently supported)` to `(only .mp4 <2GB currently supported)` in both English and Portuguese translations

2. **File validation**: Added client-side validation in `FileUploader.js` to check:
   - File type is `.mp4` (matches the existing `SUPPORTED_FILE_TYPES` constant)
   - File size is under 2GB

   If an invalid file is selected, the user is redirected to the upload page where they can see the updated restrictions.

## Related Issue

Closes #718

### DIFF
--- a/src/scripts/components/widgets/FileUploader.js
+++ b/src/scripts/components/widgets/FileUploader.js
@@ -119,6 +119,19 @@ class FilesUploader extends Component<Props, Object> {
     if (this.props.isWalletSecured) {
       const file = e.target.files[0]
       if (file) {
+        const validTypes = ['video/mp4']
+        const MAX_FILE_SIZE = 2 * 1024 * 1024 * 1024
+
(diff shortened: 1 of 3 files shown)


Karry2019web:
Superceded by PR #826 which has the same fix plus constants/validation.

PR #823, opened 2026-05-25
feat: deduplicate repeated notifications with shake animation (#751)

### LINKED ISSUE #751: Remove repeated notifications
screen shot 2018-06-29 at 11 18 59 am

There are few actions that if you do in our portal it would repeat the same notification over and over again ( maybe instead of showing it repeatedly we can bounce with the user insist in the same action that will trigger a notification).

Karry2019web:
## Description

When actions like uploading multiple files or saving wallet info trigger the same notification repeatedly, the notification stacks up and clutters the UI. Instead of showing duplicates, this PR:

1. **Deduplicates** notifications in the Redux `mapStateToProps` — if a notification with the same title + message is already showing, only the first one is kept.
2. **Adds a CSS shake animation** (`@keyframes shake`) so users get visual feedback when a notification is re-triggered: the existing notification shakes to indicate the event, rather than stacking another copy.

## Related Issue
Closes #751

## Changes
- `src/scripts/containers/NotificationContainer.js` — dedup logic using `title::message` key
- `src/scripts/components/Notification.js` — added `animateKeys` prop plumbing for future shake support
- `src/styles/app.scss` — added `@keyframes shake` animation

## Testing
- Upload triggering the same "starting upload" notice twice → only one shows, first instance shakes
- Wallet "secure your wallet" re-fires → existing notification shakes instead of duplicating

### DIFF
--- a/src/scripts/components/Notification.js
+++ b/src/scripts/components/Notification.js
@@ -5,12 +5,13 @@ import Notifications from 'react-notification-system-redux'
 import Colors from 'components/foundations/base/Colors'
 
 type Props = {
-  notifications: Array<Object>
+  notifications: Array<Object>,
+  animateKeys: Array<string>
 }
 
 class Notification extends Component<Props, Array<Object>> {
   render () {
-    const { notifications } = this.props
+    const { notifications, animateKeys } = this.props
 
     // here the reference for Style the component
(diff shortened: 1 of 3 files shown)


Karry2019web:
Closing this duplicate — superseded by #825 (this one was merged into the current implementation)

